PR #15, opened 2015-05-17
Forcehttps changes

robdimsdale:
I'm open to suggestions on where the new handlers should live. I couldn't think of an obviously better place for them, so I left them with their respective web/api handlers, deferring this decision until we need to make it in the future.

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -71,22 +71,57 @@ func Authenticate(page *agouti.Page, username, password string) {
 	page.Refresh()
 }
 
-func startATC(atcBin string, atcServerNumber uint16) (ifrit.Process, uint16) {
+func createATCCommandWithFlags(atcBin string, atcServerNumber uint16, flags map[string]string) (*exec.Cmd, uint16) {
 	atcPort := 5697 + uint16(GinkgoParallelNode()) + (atcServerNumber * 100)
 	debugPort := 6697 + uint16(GinkgoParallelNode()) + (atcServerNumber * 100)
 
-	atcCommand := exec.Command(
+	atcFlags := map[string]string{
+		"-publiclyViewable":   "true",
+		"-forceHTTPS":         "false",
+		"-webListenPort":      fmt.Sprintf("%d", atcPort),
+		"-callbacksURL":       fmt.Sprintf("http://127.0.0.1:%d", atcPort),
+		"-debugListenPort":    fmt.Sprintf("%d", debugPort),
+		"-httpUsername":       "admin",
+		"-httpHashedPassword": "$2a$04$DYaOWeQgyxTCv7QxydTP9u1KnwXWSKipC4BeTuBy.9m.IlkAdqNGG", // "password"
+		"-templates":          filepath.Join("..", "web", "templates"),
+		"-public":             filepath.Join("..", "web", "public"),
+		"-sqlDataSource":      postgresRunner.DataSourceName(),
+	}
+
+	// merge in any provided values
+	if flags != nil {
+		for k, _ := range atcFlags {
+			if flags[k] != "" {
+				atcFlags[k] = flags[k]
+			}
+		}
+	}
+
+	var flagsAsStrings []string
+	for k, v := range atcFlags {
+		// boolean flag values are treated differently;
+		// they require = when setting to false.
+		// See https://golang.org/pkg/flag/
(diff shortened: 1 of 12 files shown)


vito:
Looks great, nice job!

Unfortunately I think we'll need to do a bit more work in addition to this though. For Concourse you have to configure the ELB to do TCP or SSL forwarding, not HTTP/HTTPS, for `fly hijack` to work. So there won't be any X-Forwarded-For header. I totally forgot about that little detail when discussing this on Tracker...

Apparently there's a [proxy protocol](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-proxy-protocol.html) thing you can set up, but it might require more configuration on the backend, since it sounds like it's its own protocol *wrapping* HTTP.

I wonder if this would be simpler if we just did SSL termination in the ATC instead of relying on the ELB... That would at least make the abstraction cleaner.

Thoughts?

robdimsdale:
Yeah that makes sense. I don't think that custom proxy sounds very useful - especially given you cannot configure it through the AWS console - it's API only.

I like the simpler idea of terminating SSL on the ATC, and I think @zachgersh mentioned that golang supports TLS termination out of the box. I'm not sure of what certificates it uses for this though, and it seems that the right way to proceed with this would be to allow the operator to optionally provide a key/cert pair in the manifest if they wish, as an additional property applied only if `force_https` is true.

This solves the issue of web and APIcalls to the ATC which is the primary concern. It does, however, raise the question of whether this property should also be applied to all the traffic between all components (workers to ATC, etc) as well as external traffic inbound to the ATC. This seems like a feature that could be desirable in the enterprise world. But it also feels like a separate story.

I think we can implement the proposed solution of terminate traffic via TLS server on the ATC for web and API, using the same principles as before (reject http API, redirect web) using built-in golang support (i.e. don't allow user-provided certs at this time). Thoughts @zachgersh ?

vito:
I'd imagine you'd configure a cert directly in the manifest, yes; and `force_https` just determines whether to redirect (web) or reject (api) requests to the non-HTTP listener. I'm guessing the golang APIs require you to give it a cert anyway.

We may have to actually start using certs internally for at least TSA <-> ATC communication, since that goes over the API. Previously it just went over plaintext using the internal address, which would no longer work.

robdimsdale:
I don't really follow how hijack uses non-http connections? It looks like everything goes through a single http endpoint (https://github.com/concourse/atc/blob/master/cmd/atc/main.go#L404), and hijack is just another command in the api handler (https://github.com/concourse/atc/blob/master/api/handler.go#L100) so shouldn't hijack still work as currently implemented in this PR?

Assuming we forward raw TCP from the ELB to ATC, then one solution could be to spin up a TLS server alongside the existing server in `cmd/main.go` and use the existing API/web handlers. This would mean we would need to pull the logic of redirect/reject up out of the handlers, to a level higher (i.e. `cmd/main.go`). This is because the handlers won't know whether their connection came in on TLS or not.

zachgersh:
Sorry for the delay!

SSL termination at the ATC level seems like a promising idea.  I agree with @vito that it cleans up the abstraction, though it will require a bit more work.

So, go does have the facilities to server TLS: http://golang.org/pkg/net/http/#ListenAndServeTLS.  Will even allow us to use the handlers that have already been setup within the ATC.

@robdimsdale - As I understood your solution, you are saying boot a secondary TLS server that all http traffic will be forwarded too from `main.go`?  This is probably a fine way to go.

robdimsdale:
@vito I think I observed what you were describing - with this PR my concourse workers were unable to communicate with the ATC. I hope our proposed solution will fix this.

vito:
@robdimsdale The hijack endpoint is implemented as a regular HTTP client/server request, with one difference: both ends hijack the connection after the initial 200 OK response, and from then on it's direct TCP flowing in both directions. This doesn't work with HTTP(S) ELB configurations because those do one straight HTTP roundtrip, so what you'll see is `fly hijack` immediately exits with no output. This is also why CF typically has a separate listener for WebSockets, e.g. for routing to Loggregator.

zachgersh:
@robdimsdale maybe we close this out now?  Start this off again in a different way?

robdimsdale:
Yes we should kill this. I don't think we need to keep the branch around as reference, though I don't feel strongly if we want to keep it. Either way the PR certainly doesn't need to be open.

PR #18, opened 2015-06-10
Adds configurable check interval

zachgersh:
- All tests work :+1: 
- Moved away from channels unlike the original solution I had
- Tweaked tests to accommodate fetching the resourceConfig earlier
- branch name is still misspelled because laziness :doughnut:

### DIFF
--- a/api/configserver/save.go
+++ b/api/configserver/save.go
@@ -273,6 +273,4 @@ func sanitize(root interface{}) (interface{}, error) {
 	default:
 		return rootVal, nil
 	}
-
-	return nil, errors.New(fmt.Sprintf("unknown type (%s) during sanitization: %#v\n", reflect.TypeOf(root).String(), root))
 }
--- a/api/configserver/save.go
+++ b/api/configserver/save.go
@@ -11,6 +11,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"reflect"
+	"time"
 
 	"github.com/concourse/atc"
 	"github.com/concourse/atc/db"
@@ -151,6 +152,7 @@ func requestToConfig(contentType string, requestBody io.ReadCloser) (interface{}
 			err = yaml.Unmarshal(body, &configStructure)
 		}
 	case "multipart/form-data":
+		err = json.NewDecoder(requestBody).Decode(&configStructure)
 		multipartReader := multipart.NewReader(requestBody, params["boundary"])
 
 		for {
@@ -209,6 +211,13 @@ func saveConfigRequestUnmarshler(r *http.Request) (atc.Config, db.PipelinePaused
 			if valKind == reflect.Map {
 				if dataKind == reflect.Map {
 					return sanitize(data)
+				}
+			}
+
+			if dataKind == reflect.String {
+				val, err := time.ParseDuration(data.(string))
+				if err == nil {
+					return val, nil
 				}
 			}
 
--- a/api/configserver/save.go
+++ b/api/configserver/save.go
@@ -11,7 +11,6 @@ import (
 	"mime/multipart"
 	"net/http"
 	"reflect"
-	"time"
 
 	"github.com/concourse/atc"
 	"github.com/concourse/atc/db"
@@ -152,7 +151,6 @@ func requestToConfig(contentType string, requestBody io.ReadCloser) (interface{}
 			err = yaml.Unmarshal(body, &configStructure)
 		}
 	case "multipart/form-data":
-		err = json.NewDecoder(requestBody).Decode(&configStructure)
 		multipartReader := multipart.NewReader(requestBody, params["boundary"])
 
 		for {
@@ -211,13 +209,6 @@ func saveConfigRequestUnmarshler(r *http.Request) (atc.Config, db.PipelinePaused
 			if valKind == reflect.Map {
 				if dataKind == reflect.Map {
 					return sanitize(data)
-				}
-			}
-
-			if dataKind == reflect.String {
-				val, err := time.ParseDuration(data.(string))
-				if err == nil {
-					return val, nil
 				}
 			}
 
--- a/radar/radar.go
+++ b/radar/radar.go
@@ -197,9 +197,9 @@ func (radar *Radar) checkLock(resourceName string) []db.NamedLock {
 	return []db.NamedLock{db.ResourceCheckingLock(resourceName)}
 }
 
-func (radar *Radar) configureTimer(checkEvery atc.Duration) {
-	if checkEvery != 0 {
-		radar.interval = time.Duration(checkEvery)
+func (radar *Radar) configureTimer(checkEvery string) {
+	if checkEvery != "" {
+		radar.interval, _ = time.ParseDuration(checkEvery)
 	}
 	radar.timer = time.NewTimer(radar.interval)
 }
--- a/radar/radar.go
+++ b/radar/radar.go
@@ -48,9 +48,9 @@ type Radar struct {
 
 	interval time.Duration
 
-	locker             Locker
-	db                 RadarDB
-	updateIntervalChan chan time.Duration
+	locker Locker
+	db     RadarDB
+	timer  *time.Timer
 }
 
 func NewRadar(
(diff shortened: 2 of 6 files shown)


radar/radar.go
```diff
@@ -68,7 +68,17 @@ func NewRadar(
 
 func (radar *Radar) Scanner(logger lager.Logger, resourceName string) ifrit.Runner {
 	return ifrit.RunFunc(func(signals <-chan os.Signal, ready chan<- struct{}) error {
-		ticker := time.NewTicker(radar.interval)
+		var ticker *time.Ticker
+		resourceConfig, err := radar.getResourceConfig(logger, resourceName)
```
vito: getting this only once at the start won't cut it since they may reconfigure it at runtime, and we want to respect the new interval.

might be better to just have .scan return the interval that it found, and continuously reassign a timer using the returned interval (rather than a ticker)

zachgersh: I could get behind that though it seems a bit weird that scan now has this interval it returns.  Would you envisage that Radar has the timer directly on itself?

vito: not too worried about those lower-level details, just that mutating the object is riskier (now have to potentially deal with concurrent calls to scan, though in this case they're wrapped in a db lock). imo it's easier to keep track of return values vs. who's changing what state and having two implicitly inter-dependent code paths.

zachgersh:
Currently, the tests that fail are at these lines:
https://github.com/concourse/atc/pull/18/files#diff-885561857a7398ccbaa593fb75b0d874R373
and
https://github.com/concourse/atc/pull/18/files#diff-885561857a7398ccbaa593fb75b0d874R393

Unsure what I am doing wrong test wise here.

radar/radar.go
```diff
@@ -198,6 +211,27 @@ func (radar *Radar) checkLock(resourceName string) []db.NamedLock {
...
 }
 
+func (radar *Radar) getResourceConfig(logger lager.Logger, resourceName string) (atc.ResourceConfig, error) {
+	var found bool
+	var resourceConfig atc.ResourceConfig
+
+	config, _, err := radar.db.GetConfig()
+	if err != nil {
+		logger.Error("failed-to-get-config", err)
+		// don't propagate error; we can just retry next tick
```
vito: comment is no longer true, and i think we still don't want to propagate the error here

radar/radar.go
```diff
@@ -92,6 +108,8 @@ func (radar *Radar) Scanner(logger lager.Logger, resourceName string) ifrit.Runn
 				if err != nil {
 					return err
 				}
+
+				timer.Reset(radar.interval)
```
vito: I don't think a timer can be reset after it's already fired.

radar/radar.go
```diff
@@ -77,7 +90,10 @@ func (radar *Radar) Scanner(logger lager.Logger, resourceName string) ifrit.Runn
 			case <-signals:
 				return nil
 
-			case <-ticker.C:
+			case newInterval := <-radar.updateIntervalChan:
```
vito: channel seems a bit weird for this since it's always a synchronous function call that's populating it (which is why it has to be buffered, otherwise it'd deadlock)

radar/radar.go
zachgersh: @vito - you are totally right.  I thought you were crazy but I went back and read the docs again.  Don't know how I missed that :/

radar/radar.go
zachgersh: @vito paired with your comment above I actually think the channel is going to work out nicely.  I am going to make it so that a brand new timer with the correct interval is sent down the channel instead.

I like this approach a lot better than making scan return an interval when it doesn't really have a reason to.

radar/radar.go
zachgersh: Great catch.  I will take this out.

xoebus:
@zachgersh Reviewed this today and put some notes on the story: https://www.pivotaltracker.com/story/show/76230000

zachgersh:
@xoebus - I thought using an atc.Duration took care of this?

https://github.com/concourse/atc/blob/master/config.go#L289

xoebus:
Hmmm, I thought so too - I'll take another look.

zachgersh:
@xoebus this should now be ready to test again!

api/configserver/save.go
```diff
@@ -151,6 +152,7 @@ func requestToConfig(contentType string, requestBody io.ReadCloser) (interface{}
 			err = yaml.Unmarshal(body, &configStructure)
 		}
 	case "multipart/form-data":
+		err = json.NewDecoder(requestBody).Decode(&configStructure)
```
vito: what's this for?

zachgersh: Cruft, I will pull that out.

radar/radar.go
```diff
@@ -120,28 +126,21 @@ func (radar *Radar) scan(logger lager.Logger, resourceName string) error {
...
-	savedResource, err := radar.db.GetResource(resourceName)
+	resourceConfig, err := radar.getResourceConfig(logger, resourceName)
 	if err != nil {
 		return err
 	}
 
-	if savedResource.Paused {
-		return nil
-	}
+	defer radar.configureTimer(resourceConfig.CheckEvery)
```
vito: might as well just do this immediately; don't think the `defer` is needed

api/configserver/save.go
```diff
@@ -212,6 +214,13 @@ func saveConfigRequestUnmarshler(r *http.Request) (atc.Config, db.PipelinePaused
 				}
 			}
 
+			if dataKind == reflect.String {
```
vito: is there any way to do this only if the valKind expects a `time.Duration` / `atc.Duration`?

vito:
only a few minor nits, looking pretty good!

radar/radar.go
zachgersh: @vito I defer so that I don't reconfigure the timer until the whole function has finished.  Are you sure you want the behavior where the timer gets created and starts before that initial check has finished?

vito: ah makes sense, forgot they start counting immediately

radar/radar.go
```diff
@@ -68,7 +69,23 @@ func NewRadar(
 
 func (radar *Radar) Scanner(logger lager.Logger, resourceName string) ifrit.Runner {
 	return ifrit.RunFunc(func(signals <-chan os.Signal, ready chan<- struct{}) error {
-		ticker := time.NewTicker(radar.interval)
+		resourceConfig, err := radar.getResourceConfig(logger, resourceName)
+		if err != nil {
+			return err
+		}
+
+		err = func() (err error) {
```
vito: what's up with this?

radar/radar.go
```diff
@@ -108,47 +125,48 @@ func (radar *Radar) Scan(logger lager.Logger, resourceName string) error {
 	return radar.scan(logger, resourceName)
 }
 
-func (radar *Radar) scan(logger lager.Logger, resourceName string) error {
+func (radar *Radar) scan(logger lager.Logger, resourceName string) (err error) {
```
vito: we've been avoiding named return values

radar/radar.go
```diff
@@ -108,47 +125,48 @@ func (radar *Radar) Scan(logger lager.Logger, resourceName string) error {
...
+		return
 	}
 
-	config, _, err := radar.db.GetConfig()
+	savedResource, err := radar.db.GetResource(resourceName)
 	if err != nil {
-		logger.Error("failed-to-get-config", err)
-		// don't propagate error; we can just retry next tick
-		return nil
+		return
```
vito: bug here introduced by named returns: err return value should be nil. or did you mean for this change?

radar/radar.go
```diff
@@ -108,47 +125,48 @@ func (radar *Radar) Scan(logger lager.Logger, resourceName string) error {
...
-		return err
+		return
 	}
 
-	if savedResource.Paused {
-		return nil
-	}
+	defer func() {
+		if resourceConfig.CheckEvery != "" {
+			radar.interval, err = time.ParseDuration(resourceConfig.CheckEvery)
```
vito: guessing this (overriding `err`) is why you switched to named returns, but it seems tricky/hard to notice. better to be explicit here imo

zachgersh: @vito - yup you totally nailed it and that was the thing I would worried about (that it would be tricky to parse out).  So it's named returns all the way down now.  Since this code has to be deferred there isn't a really good way to get values out of any function that is deferred unless the variables are pre-declared in the outer scope.

I am thinking this may be a bit of necessary complexity?  I could also add a comment to point out the fact that we've gone to named returns.

zachgersh:
@vito this should be good now.  I went backwards and decided that if we fail parsing the interval we should just use the one that was previously set (instead of throwing an error).  This allowed me to revert the named returns.

I think this is probably the best approach - we shouldn't need to error out if we already had a previously working interval.

cc @xoebus

PR #50, opened 2016-02-17
Attempt to create container on another worker if an error occurs

mdelillo:
MicroPCF team requires this in order to limit the number of containers in our custom Houdini workers.

We think this is a generally useful behavior. Failed attempts to create containers are logged.

### DIFF
--- a/exec/task_step.go
+++ b/exec/task_step.go
@@ -3,6 +3,7 @@ package exec
 import (
 	"archive/tar"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


davidwadden:
@vito Any progress on getting this merged?  We'd like this merged in so that we can upgrade our Concourse installation without having to apply the patch and re-deploy.

@davidwadden
PCF Dev PM

zachgersh:
@davidwadden seems like your branch has merge conflicts now?

sclevine:
@zachgersh We were waiting for confirmation that this would get merged before we rebased. We'll rebase now to speed things along.

vito:
We're not likely to merge this, at least any time soon. We're shooting for a 1.0 release soon and I'm certain we won't have enough time to characterize the resulting behavior of this change. I'm also not sure it's the right way about it; I'd rather have the ATC schedule things a bit more intelligently based on the capacity reported by the workers.

sclevine:
@vito Our CI is heavily reliant on this change in behavior. We were told in Slack that this implementation would be reasonable, and we put off upgrading our concourse deployment for the last month while waiting for this PR to get merged. Could we introduce a flag and corresponding BOSH property to make this behavior opt-in? If not, we can continue to patch our ATC, but it makes upgrading concourse significantly more painful.

vito:
@sclevine I remember asking if it could check for specific capacity-related errors rather than retrying on everything (potentially exacerbating an issue across all workers), which is part of my concern with the current state of things. I don't remember explicitly saying we approve of this approach, but I guess we never said no either.

I'll think about the flag but, again, I'd rather not buy too much into this (and support it, and document it, and [...]) given that I'd prefer alternative approaches like explicit smarter scheduling.

Again sorry for the radio silence.

sclevine:
@vito What if we limit the change in behavior to a specific, `atc`-typed error that only our modified Houdini backend would return? This would result in zero changes in behavior or logic for anyone except us. Other backends could opt-in to this logic in the future, or more robust scheduling could be implemented in ATC.

sclevine:
@vito What do you think of the proposition above? We extracted a large method, but our change is actually very small and will have no effect on backends that don't return the typed error (if we make the change).

This is the third time we've had to rebase this PR. It would be nice to know how we should proceed.

Thanks!

(CC @davidwadden)

flavorjones:
@vito Just pinging about @sclevine's proposal from last week. Is a typed error something you'd consider merging? I'd just like to be able to set expectations, whether the answer is yes or (presumably) no.

vito:
@flavorjones @sclevine We won't be merging this, especially with 1.0 on its way. Sorry for the silence; been focusing on shipping 1.0 and getting the docs in order; it wasn't until recently that we decided to push it out farther.

I don't like the idea of modifying production code to look for an error that only one Concourse deployment will ever return; a better approach would be to improve container placement in the first place.

For now I'd continue rebasing, unless you want to sink time into that latter approach.

Closing for clarity.

sclevine:
Can we re-open this now that 1.0 has shipped?

@vito I put a meeting on your calendar to discuss -- I'd like to come to an agreement on an iteration of this that you would accept.

PR #53, opened 2016-02-19
Set identifying info for check container

zachgersh:
This addresses:

https://github.com/concourse/concourse/issues/258

Hoping that passing external URL around like this is ok. Additionally, tried to pick criteria that would identify a check container - it may be wrong.

Where should the documentation for this feature go?

Signed-off-by: Utako Ueda <uueda@pivotal.io>

### DIFF
--- a/api/jobserver/create_build.go
+++ b/api/jobserver/create_build.go
@@ -33,7 +33,7 @@ func (s *Server) CreateJobBuild(pipelineDB db.PipelineDB) http.Handler {
 			return
 		}
 
-		scheduler := s.schedulerFactory.BuildScheduler(pipelineDB)
+		scheduler := s.schedulerFactory.BuildScheduler(pipelineDB, s.externalURL)
 
 		build, _, err := scheduler.TriggerImmediately(logger, job, config.Resources, config.ResourceTypes)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


robdimsdale:
@vito @xoebus any thoughts on this PR? What about the question @zachgersh asked - where should documentation for this PR go?

zachgersh:
@vito - any other changes needed here?

xoebus:
No, looks great. Thanks!

vito:
Looking at this again, it's inconsistent with the metadata provided to `/in` and `/out`:

* `BUILD_PIPELINE_NAME` vs. `PIPELINE_NAME`: perhaps it should be `RESOURCE_PIPELINE_NAME`?
* `EXTERNAL_URL` vs. `ATC_EXTERNAL_URL`: this should be `ATC_EXTERNAL_URL`.

vito:
Gonna leave this undocumented for now til that's fixed. /cc @robdimsdale

robdimsdale:
@zachgersh if/when you update these environment variables here, can you also update the `check` user-agent in the pivnet-resource to respect these changes?

zachgersh:
@robdimsdale will do :dancers:

Issue #35, opened 2015-10-29
Sidebar doesen't scroll


roxtar:
![screen shot 2015-10-29 at 11 21 31 am](https://cloud.githubusercontent.com/assets/128514/10826413/59e6bdc2-7e2f-11e5-801f-3bc5cef5ce6f.png)

I can't scroll down to the pipelines further down on the sidebar.

PR #63, opened 2016-03-09
Scroll pipeline navigation

jhunt:
Installations that have lots of pipelines eventually run into issues
scrolling to find their pipeline in the list.  This commit fixes that.

### DIFF
--- a/web/assets/css/pipeline_nav.less
+++ b/web/assets/css/pipeline_nav.less
@@ -80,6 +80,7 @@
 }
 
 .pipelinesNav-enabled .pipelines li {
+  position: relative;
   padding: 0;
   margin: 0;
   text-align: right;
@@ -100,6 +101,12 @@
     text-overflow: ellipsis;
     white-space: nowrap;
     overflow: hidden;
+  }
+
+  span.fl {
+    position: absolute;
+    top: 0px;
+    left: 0px;
   }
 }
 
--- a/web/assets/css/pipeline_nav.less
+++ b/web/assets/css/pipeline_nav.less
@@ -55,6 +55,8 @@
 
   position: fixed;
   top: 40px;
+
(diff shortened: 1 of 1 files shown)


jhunt:
I believe this fixes #35

vito:
sweet. looks a little janky here though, probably from the additional space taken up by the sidebar.

screenshot 2016-03-09 at 12 30 43 pm

jhunt:
@vito I can't repro that.  What browser are you on, and is this Concourse publicly accessible?

vito:
@jhunt chrome on a chromebook pixel. that one's not publicly accessible but you could just resize the window on https://ci.concourse.ci - it happens there too

screenshot 2016-03-09 at 12 43 34 pm

jhunt:
Still no joy on repro.  Tried FF / Safari / Chrome on OS X El Cap.  Here's Chrome:

screen shot 2016-03-09 at 3 50 08 pm

vito:
probably due to the OS x overlay scrollbars. you could turn that off in sys
prefs

jhunt:
That worked.  I'll dig into that and see what I can come up with

jhunt:
Seems to clear up if I set the following:

```
.pipelinesNav-enabled .pipelines li         { position: relative; }
.pipelinesNav-enabled .pipelines li span.fl { position: absolute; top: 0px; left: 0px; }
```

To take the button (`span.fl`) out of the document flow without adversely affecting its positioning.  Thoughts?

vito:
Yup that fixes it. There may be a more elegant way to do it but I'm not too worried about it; we'll probably be redoing the sidebar the next time we touch it ourselves.

vito:
Merging, thanks! I'll rebuild the .css locally and push so it takes effect.

PR #65, opened 2016-03-14
Update External URL and Pipeline Name

utako:
Addresses #53
* For consistency

[#113219413]

Signed-off-by: Zachary Gershman <zgershman@pivotal.io>

PR #76, opened 2016-04-25
Remove need for auth on cli download

zachgersh:
Removed the need to be authenticated when downloading the CLI (even when not public). This should pretty much always be a method that you can get to on any concourse server.

### DIFF
--- a/wrappa/api_auth_wrappa.go
+++ b/wrappa/api_auth_wrappa.go
@@ -68,11 +68,12 @@ func (wrappa *APIAuthWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
 			newHandler = auth.CheckAuthHandler(handler, rejector)
 
 		// unauthenticated
-		case atc.ListAuthMethods, atc.GetInfo:
+		case atc.ListAuthMethods,
+			atc.GetInfo,
+			atc.DownloadCLI:
 
 		// unauthenticated if publicly viewable
 		case atc.BuildEvents,
-			atc.DownloadCLI,
 			atc.GetBuild,
 			atc.GetJobBuild,
 			atc.BuildResources,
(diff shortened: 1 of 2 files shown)


krishicks:
Hi @zachgersh,

This change would only impact configurations where the atc is entirely locked down, i.e. not publicly viewable. It seems weird to us to let you make this request unauthenticated if the atc has been locked down.

Can you clarify where the need to download the cli when not authenticated comes from?

zachgersh:
Hey @krishicks,

Thanks for the reply! So, issue #65 above your message is the problem that happens when this isn't allowed to be downloaded un-authenticated (the dreaded sync/login loop). 

This CLI endpoint doesn't expose any information about your concourse installation - so I don't really see a reason for it to actually be locked down (plus we have two other endpoints already exposed when you are un-authenticated). I'd love to understand what locking down the cli endpoint actually bought the concourse installation?

vito:
@zachgersh We fixed the login loop already in `fly`. The other two endpoints have to be permitted for you be able to log in *at all*. The CLI isn't as necessary.

The point being if you haven't made your pipeline publicly viewable you likely want it as locked-down as possible.

krishicks:
@zachgersh Is it OK for us to close this, given @vito's response above?

zachgersh:
@krishicks / @vito - I'd actually like to see this merged. I didn't even consider that fact that you could be trying to download the CLI programmatically and forcing people to login to do that seems like a poor user experience as well.

vito:
@zachgersh Again, if the whole instance is locked down I'd want it to be locked down as much as possible. You can still script downloads with basic auth by acquiring a token first via the `/api/v1/auth/token` endpoint.

vito:
Closed this as the original issue in Fly has already been fixed, and the related issue (#65) has been resolved.

PR #77, opened 2016-04-26
jobs API endpoint serves SVG badges of latest status

### LINKED ISSUE #26: A job with complicated `passed` constraints takes a very long time to determine its inputs
Today we execute a pretty expensive query to determine the candidate versions for a build's `get` steps, based on their `passed` constraints. If the constraints are complicated enough (either `passed: [a0, a1, ... aN]` or many inputs with correlated passed constraints), this kills the postgres as more and more versions and more and more builds have to be scanned through.

The related database logic is [`(PipelineDB).GetLatestInputVersions`](
https://github.com/concourse/atc/blob/ab1071474b8fb8f0b452d273961f75e88df4c9ff/db/pipeline_db.go#L1373). It already attempts some optimization (by splitting up the query to one per input, adding constraints as it goes along), but it's still very expensive.

cunnie:
@xoebus did the heavy lifting.

- typical URI:
  [https://ci.blabbertabber.com/api/v1/pipelines/main/jobs/always-succeeds/badge](https://ci.blabbertabber.com/api/v1/pipelines/main/jobs/always-succeeds/badge)

| Build Status      | Shield                                                                 | Implemented |
--------------------|------------------------------------------------------------------------|----------------|
| `nil` (build never run)             | unknown   | yes |
| `StatusAborted`   | unknown   | yes |
| `StatusSucceeded` | passing | yes |
| `StatusFailed`    | failing         | yes |
| `StatusErrored`   | unknown   | yes |

- Similar to the badges served by [Travis CI](https://travis-ci.com/)
- SVG created courtesy [shields.io](http://shields.io/)
- Badge style: flat

Here's a deployment of a Concourse dev release with the badges: [https://ci.blabbertabber.com](https://ci.blabbertabber.com):

badge-pipeline

And here are the badges (***pulled from the live site!***):

- passing
- failing
- aborted
- errored
- unknown

[fixes #72388008]

Signed-off-by: Brian Cunnie <bcunnie@pivotal.io>

### DIFF
--- a/api/jobs_test.go
+++ b/api/jobs_test.go
@@ -260,6 +260,291 @@ var _ = Describe("Jobs API", func() {
 		})
 	})
 
+	Describe("GET /api/v1/pipelines/:pipeline_name/jobs/:job_name/badge", func() {
+		var response *http.Response
+
+		JustBeforeEach(func() {
+			var err error
+
+			response, err = client.Get(server.URL + "/api/v1/pipelines/some-pipeline/jobs/some-job/badge")
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(pipelineDBFactory.BuildWithTeamNameAndNameCallCount()).To(Equal(1))
+			teamName, pipelineName := pipelineDBFactory.BuildWithTeamNameAndNameArgsForCall(0)
+			Expect(pipelineName).To(Equal("some-pipeline"))
+			Expect(teamName).To(Equal(atc.DefaultTeamName))
+		})
+
+		Context("when getting the job config succeeds", func() {
+			BeforeEach(func() {
+				pipelineDB.GetConfigReturns(atc.Config{
+					Groups: []atc.GroupConfig{
+						{
+							Name: "group-1",
+							Jobs: []string{"some-job"},
+						},
+						{
+							Name: "group-2",
+							Jobs: []string{"some-job"},
+						},
+					},
+
+					Jobs: []atc.JobConfig{
+						{
+							Name: "some-job",
+							Plan: atc.PlanSequence{
+								{
+									Get: "some-input",
+								},
+								{
+									Get:      "some-name",
+									Resource: "some-other-input",
+									Params:   atc.Params{"secret": "params"},
+									Passed:   []string{"a", "b"},
+									Trigger:  true,
+								},
+								{
+									Put: "some-output",
+								},
+								{
+									Put:    "some-other-output",
+									Params: atc.Params{"secret": "params"},
+								},
+							},
+						},
+					},
+				}, 1, true, nil)
+			})
+
+			Context("when the finished build is successful", func() {
+				BeforeEach(func() {
+					pipelineDB.GetJobFinishedAndNextBuildReturns(
+						&db.Build{
+							ID:           1,
+							Name:         "1",
+							JobName:      "some-job",
+							PipelineName: "some-pipeline",
+							Status:       db.StatusSucceeded,
+							StartTime:    time.Unix(1, 0),
+							EndTime:      time.Unix(100, 0),
+						},
+						&db.Build{
+							ID:           3,
+							Name:         "2",
+							JobName:      "some-job",
+							PipelineName: "some-pipeline",
+							Status:       db.StatusStarted,
+						},
+						nil,
+					)
+				})
+
+				It("fetches by job", func() {
+					Expect(pipelineDB.GetJobFinishedAndNextBuildCallCount()).To(Equal(1))
+
+					jobName := pipelineDB.GetJobFinishedAndNextBuildArgsForCall(0)
+					Expect(jobName).To(Equal("some-job"))
+				})
+
+				It("returns 200 OK", func() {
+					Expect(response.StatusCode).To(Equal(http.StatusOK))
+				})
+
+				It("returns some SVG showing that the job is successful", func() {
+					body, err := ioutil.ReadAll(response.Body)
+					Expect(err).NotTo(HaveOccurred())
+
+					Expect(string(body)).To(Equal(`<svg xmlns="http://www.w3.org/2000/svg" width="88" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="88" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h37v20H0z"/><path fill="#44cc11" d="M37 0h51v20H37z"/><path fill="url(#b)" d="M0 0h88v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="18.5" y="15" fill="#010101" fill-opacity=".3">build</text><text x="18.5" y="14">build</text><text x="61.5" y="15" fill="#010101" fill-opacity=".3">passing</text><text x="61.5" y="14">passing</text></g></svg>`))
+				})
+			})
+
+			Context("when the finished build is failed", func() {
+				BeforeEach(func() {
+					pipelineDB.GetJobFinishedAndNextBuildReturns(
+						&db.Build{
+							ID:           1,
+							Name:         "1",
+							JobName:      "some-job",
+							PipelineName: "some-pipeline",
+							Status:       db.StatusFailed,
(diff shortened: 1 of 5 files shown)


api/jobserver/badge.go
```diff
@@ -0,0 +1,74 @@
...
+		if finished.Status == db.StatusFailed {
+			fmt.Fprint(w, getSvg(badgeFailing))
+			return
+		}
+	})
+}
+
+func getSvg(badge badge) string {
+	const svgTemplate = `<svg xmlns="http://www.w3.org/2000/svg" width="%d" height="20"><linearGradient id="b" x2="0" y2="100%%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="%d" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h37v20H0z"/><path fill="%s" d="M37 0h%dv20H37z"/><path fill="url(#b)" d="M0 0h%dv20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="18.5" y="15" fill="#010101" fill-opacity=".3">build</text><text x="18.5" y="14">build</text><text x="%.1f" y="15" fill="#010101" fill-opacity=".3">%s</text><text x="%.1f" y="14">%s</text></g></svg>`
+	return fmt.Sprintf(svgTemplate, badge.width, badge.width, badge.fillColor, badge.width-37, badge.width, float64(badge.width)/2+17.5, badge.status, float64(badge.width)/2+17.5, badge.status)
```
xoebus: Nice work working out these values!

Maybe extract the `badge.width-37` and other computations as methods on the `badge` struct e.g. `badge.padding()` This would make them a little less magic by giving them names.

vito:
looks great! prioritizing.

api/jobserver/badge.go
```diff
@@ -0,0 +1,96 @@
...
+			fmt.Fprint(w, badgeAborted.getSvg())
+			return
+		}
+
+		if finished.Status == db.StatusErrored {
+			fmt.Fprint(w, badgeErrored.getSvg())
+			return
+		}
+
+		fmt.Fprint(w, badgeUnknown.getSvg())
```
xoebus: Maybe this block should be switched to a `switch` statement?

api/jobserver/badge.go
```diff
@@ -0,0 +1,96 @@
...
+
+		w.Header().Set("Content-type", "image/svg+xml")
+
+		w.WriteHeader(http.StatusOK)
+
+		badgePassing := badge{width: 88, fillColor: `#4c1`, status: `passing`}
+		badgeFailing := badge{width: 80, fillColor: `#e05d44`, status: `failing`}
+		badgeUnknown := badge{width: 98, fillColor: `#9f9f9f`, status: `unknown`}
+		badgeAborted := badge{width: 90, fillColor: `#8f4b2d`, status: `aborted`}
+		badgeErrored := badge{width: 88, fillColor: `#fe7d37`, status: `errored`}
```
xoebus: These can be extracted as package variables.

ernado:
I need it so bad, looks great.
Will use it definitely :)

api/jobserver/badge.go
ernado: Maybe something like this?
```go
var (
	badgePassing = badge{width: 88, fillColor: `#4c1`, status: `passing`}
	badgeFailing = badge{width: 80, fillColor: `#e05d44`, status: `failing`}
	badgeUnknown = badge{width: 98, fillColor: `#9f9f9f`, status: `unknown`}
	badgeAborted = badge{width: 90, fillColor: `#8f4b2d`, status: `aborted`}
	badgeErrored = badge{width: 88, fillColor: `#fe7d37`, status: `errored`}
)

func getBadge(b *db.Build) badge {
	if b == nil {
		return badgeUnknown
	}
	switch b.Status {
	case db.StatusSucceeded:
		return badgePassing
	case db.StatusFailed:
		return badgeFailing
	case db.StatusAborted:
		return badgeAborted
	case db.StatusErrored:
		return badgeErrored
	default:
		return badgeUnknown
	}
}

// and in JobBadge
// fmt.Fprint(w, getBadge(finished).getSvg())
// or more verbose
```

api/jobs_test.go
```diff
@@ -260,6 +260,291 @@ var _ = Describe("Jobs API", func() {
...
+	Describe("GET /api/v1/pipelines/:pipeline_name/jobs/:job_name/badge", func() {
+		var response *http.Response
+
+		JustBeforeEach(func() {
+			var err error
+
+			response, err = client.Get(server.URL + "/api/v1/pipelines/some-pipeline/jobs/some-job/badge")
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(pipelineDBFactory.BuildWithTeamNameAndNameCallCount()).To(Equal(1))
```
krishicks: Could these expectations be moved out of the JustBeforeEach? It seems like they're actual test expectations rather than setup.

api/jobs_test.go
```diff
@@ -260,6 +260,291 @@ var _ = Describe("Jobs API", func() {
...
+							Name:         "2",
+							JobName:      "some-job",
+							PipelineName: "some-pipeline",
+							Status:       db.StatusStarted,
+						},
+						nil,
+					)
+				})
+
+				It("fetches by job", func() {
```
krishicks: This and the duplicate `It` should be moved up to line 318; GetJobFinishedAndNextBuild being called should be asserted on once; the following contexts ("when it's successful, when it is failed") then set up what happens afterwards.

api/jobs_test.go
```diff
@@ -260,6 +260,291 @@ var _ = Describe("Jobs API", func() {
...
+					}, 1, true, nil)
+				})
+
+				It("returns 404", func() {
+					Expect(response.StatusCode).To(Equal(http.StatusNotFound))
+				})
+			})
+		})
+
+		Context("when getting the job config fails", func() {
```
krishicks: I'd remove this empty `Context`. It has no `BeforeEach`, and doesn't really add anything.

api/jobserver/badge.go
```diff
@@ -0,0 +1,96 @@
...
+func (b *badge) statusWidth() int {
+	return b.width - 37
+}
+
+func (b *badge) statusTextWidth() float64 {
+	return float64(b.width)/2 + 17.5
+}
+
+func (b *badge) getSvg() string {
+	const svgTemplate = `<svg xmlns="http://www.w3.org/2000/svg" width="%d" height="20"><linearGradient id="b" x2="0" y2="100%%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="%d" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h37v20H0z"/><path fill="%s" d="M37 0h%dv20H37z"/><path fill="url(#b)" d="M0 0h%dv20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="18.5" y="15" fill="#010101" fill-opacity=".3">build</text><text x="18.5" y="14">build</text><text x="%.1f" y="15" fill="#010101" fill-opacity=".3">%s</text><text x="%.1f" y="14">%s</text></g></svg>`
```
krishicks: This should be converted into a proper template using the `template` package. https://golang.org/pkg/text/template/

It would be much nicer if this template was also formatted with newlines. Adding/modifying this template as-is would be very difficult because you'd have to count the number of '%d' and '%s' values. Instead, using the `template` package you can provide a struct that has as its members the fields to be used to fill out the template.

krishicks: e.g.
```
<svg xmlns="http://www.w3.org/2000/svg" width="{{ .Width }}" height="20">
   <linearGradient id="b" x2="0" y2="100%%">
      <stop offset="0" stop-color="#bbb" stop-opacity=".1"/>
      <stop offset="1" stop-opacity=".1"/>
   </linearGradient>
   <mask id="a">
      <rect width="{{ .Width }}" height="20" rx="3" fill="#fff"/>
...
```

api/jobserver/badge.go
cunnie: Thanks, working on changes now. Will push when done.

api/jobserver/badge.go
cunnie: Thanks Kris, I'll implement template package during lunch and in the evening.

PR #83, opened 2016-05-27
add a disable-intercept option

fmy:
## what
I created a new option `--disable-intercept`.
If true, `fly hijack ( or intercept)` commands will be denied.

## why
I'm considering to collaborate the concourse with our GitHub Enterprise.
I try to use a `concourse` bot user to pull/push from all pipelines (with customizing git-resource).
With this, every developers needs not to expose their private keys.
Customizing git-resource will include `concourse` private key, so I use this disable-intercept option in order to hide this private key.

Let me know if anything looks bad.

### DIFF
--- a/api/containerserver/hijack.go
+++ b/api/containerserver/hijack.go
@@ -17,6 +17,11 @@ var upgrader = websocket.Upgrader{
 }
 
 func (s *Server) HijackContainer(w http.ResponseWriter, r *http.Request) {
+	if s.disableHijack {
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
+
 	handle := r.FormValue(":id")
 
 	hLog := s.logger.Session("hijack", lager.Data{
--- a/api/containerserver/server.go
+++ b/api/containerserver/server.go
@@ -12,6 +12,8 @@ type Server struct {
 	workerClient worker.Client
 
 	db ContainerDB
+
+	disableHijack bool
 }
 
 //go:generate counterfeiter . ContainerDB
@@ -25,10 +27,12 @@ func NewServer(
 	logger lager.Logger,
 	workerClient worker.Client,
 	db ContainerDB,
+	disableHijack bool,
 ) *Server {
 	return &Server{
-		logger:       logger,
-		workerClient: workerClient,
-		db:           db,
+		logger:        logger,
+		workerClient:  workerClient,
+		db:            db,
+		disableHijack: disableHijack,
 	}
 }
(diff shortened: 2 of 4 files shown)


Pirolf:
Disabling `fly intercept` is not enough to prevent users of fly from seeing the credentials in the pipelines configured by the bot. For example `fly get-pipeline` will print out the entire pipeline config including credentials. I would like more details about your situation. The way I understand your situation, it sounds like your main goal is to share your concourse deployment with people who shouldn't have access to your repos. What exactly should they be allowed to do with your concourse? Anything beyond triggering builds? Is the bot's purpose to allow those people to set pipelines without seeing the credentials?

fmy:
> Disabling fly intercept is not enough to prevent users of fly from seeing the credentials in the pipelines configured by the bot. For example fly get-pipeline will print out the entire pipeline config including credentials. 
Yes, I think adding personal credential into pipeline is not good way. So I created a bot github user.

The way I understand your situation, it sounds like your main goal is to share your concourse deployment with people who shouldn't have access to your repos. What exactly should they be allowed to do with your concourse? Anything beyond triggering builds?
> Is the bot's purpose to allow those people to set pipelines without seeing the credentials?
Yes

fmy:
> Disabling fly intercept is not enough to prevent users of fly from seeing the credentials in the pipelines configured by the bot. For example fly get-pipeline will print out the entire pipeline config including credentials.

Yes, in my opinion, adding credentials into pipelines is not good way.
Many plain text creds would be stored in the postgres.
My teammate could steal my creds with `fly get-pipeline`.

Instead, we use one bot github user whose credential is placed into docker image resource.

evanshort73:
I see, in that case this pull request would work. Your use case seems convincing, so I will have a conversation with the concourse team to decide if this is the right way to accomplish it.

krishicks:
Future [Vault](https://www.hashicorp.com/blog/vault.html) support will mitigate the credentials-in-pipeline problem.

evanshort73:
After some more consideration, I realized that malicious team members could still run one-off builds telling your bot to do whatever they want. Therefore the only way the bot provides security is by limiting which git commands can be run. You can probably achieve the same security by creating [deploy keys] (https://developer.github.com/guides/managing-deploy-keys/) with the right permissions.

fmy:
@krishicks @evanshort73 Thanks for your advice! I will consider to use deploy keys and vault combination.
But If users will be able to fetch credentials from the vault while `hijack`ing, `disable-hijack` option still has a point.

evanshort73:
I agree that disable-hijack still has a point. Ideally I would like to implement this sort of feature on a per-user basis using OAuth scopes. We'll revisit this idea once we start working on Vault integration.

PR #85, opened 2016-06-02
fix #59 show error message when server dies

### LINKED ISSUE #59: Concourse UI should indicate that the connection to the server is down
If you are using Concourse on a wall display and the server dies, it will continue to show the same view. Build animations continue so there is no way to tell that it will never update.

I suggest opening a modal that says "Communication with the server has been interrupted" when the server connection is down.

fmy:
Showing an error message when server dies.
Please refer to #59 .
When the concourse server comes back, this error message automatically disappears.


shinji62:
Nice !

vito:
thanks! made some changes to it to fit the aerospace metaphor a bit better :P

PR #92, opened 2016-07-09
Update CONTRIBUTING.md with note about fly CLI

philwebb:
The `fly` CLI can fail with the build version. This just adds some additional notes about how to build it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,3 +81,23 @@ go run cmd/atc/*.go -d
 ```
 
 Concourse should be live at http://localhost:8080
+
+
+## Building the fly command line tool
+
+If you try to use a realse version of `fly` you will get errors like this:
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -81,3 +81,23 @@ go run cmd/atc/*.go -d
 ```
 
 Concourse should be live at http://localhost:8080
+
+
+## Building the fly command line tool
+
+If you try to use a realse version of `fly` you will get errors like this:
```
vito: This should say something like "If you try to use a released version of `fly` against a development version of `atc`, you will get the following error:"

philwebb: Want another PR or can you add that during the merge?

vito: No worries, I can add it._

PR #93, opened 2016-07-11
Reduce CPU usage for CSS animations

philwebb:
I noticed when using concourse on my laptop that viewing any pipeline screen makes it run hot, typically maxing out a single core and eating battery. After a bit of investigation it appears that the ripple animations are the primary cause. For example, an empty pipeline that only shows the legend will consume 50% CPU. If I delete the legend CPU drops back to normal.

This commit attempts to reduce CPU usage by created the ripple effect using only CSS transforms. This [article by Paul Irish](http://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/) provides more background on why this helps. The "show paint rectangles" devtools tip is particularly helpful in seeing what's going on.

Unfortunately I don't have enough pipelines defined to be totally confident that all animation sizes will animate correctly. My simple pipelines seem to work well and consume considerably less CPU. I imagine further tweaks to the CSS could be made if other sizes don't render correctly.

I've split this PR into two commits to help reduce the noise caused by regenerating assets. Only the first commit is really relevant for discussion.


vito:
Doesn't quite work in Chrome. :)

screenshot 2016-07-11 at 5 27 16 pm

philwebb:
I hate web development! Will try again.

vito:
Have I got the pipeline for you: https://ci.concourse.ci/pipelines/web-development

philwebb:
I'm afraid I'm a bit stumped. I think [this](http://stackoverflow.com/questions/15194313/webkit-css-transform3d-position-fixed-issue) may be the root cause but I'm stuck on how to fix it. Perhaps someone with more CSS skill can take the ideas and make them work with Chrome. It's quite frustrating because the improvement on my local machine was quite significant.

vito:
I'm gonna just close this for now - we're tinkering with a new pipeline UI engine that avoids using SVG, as its behavior (both performance and appearance) tends to be unpredictable from browser to browser and machine to machine.

Thanks for looking into it, though!

philwebb:
Sounds good. Hopefully the [Paul Irish video](http://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/) will be relevant with whatever technique is used to provide the next gen animations.

philwebb:
@vito Is there an issue for the new UI engine or did it get parked in the end? If the SVG animations are remaining perhaps it's worth reopening this one?

PR #98, opened 2016-08-09
Fixing ELM version from 1.6 to 0.17

shinji62:
Seems now Concourse use elm 0.17 so I just updated contributing.md to reflect it.


vito:
Ah that should really be 0.17, not 1.7 - looks like it was wrong before too. Mind fixing that along the way?

shinji62:
Will fix that :p

shinji62:
Done.

vito:
thanks, looks good!

PR #99, opened 2016-08-11
Added generic oauth login option to teams

poida:
Implements a generic OAuth provider as mentioned in this issue: https://github.com/concourse/concourse/issues/539.

Use it by specifying the following on the command line:
```
concourse
  --oauth-client-id=abc \
  --oauth-client-secret=123 \
  --oauth-auth-url=https://github.com/login/outh/authorize \
  --oauth-token-url=https://github.com/login/oauth/access_token \
  --oauth-display-name=Githubz
```

This will present a login prompt on the web ui like:
image


poida:
I realise that I'm missing *ahem* tests and documentation, but would appreciate feedback on my approach (code, variables etc).

vito:
Looks good! (Aside from the no tests thing.)

May want to rename the flags to something like `--generic-oauth` or something though; it looks a bit too "normal" and I worry people may think they should use it even when more specialized providers are available (e.g. Google or GitHub). Something to call out that it does no access verification may also be in order; maybe in the help-text?

poida:
Messed up pull request and will need to resubmit.

PR #101, opened 2016-08-15
Added generic oauth login option.

poida:
(following on from https://github.com/concourse/atc/pull/99)

The option allows all authorized users from the given auth provider. See https://github.com/concourse/concourse/issues/539.

Use it by specifying the following on the command line:
```
concourse
  --generic-oauth-client-id=abc \
  --generic-oauth-client-secret=123 \
  --generic-oauth-auth-url=https://github.com/login/outh/authorize \
  --generic-oauth-auth-param=param1:value1 \
  --generic-oauth-token-url=https://github.com/login/oauth/access_token \
  --generic-oauth-display-name=Githubz
```

This will present a login prompt on the web ui like:
image

I've verified that it works with GitHub.

Added support for a configurable display name.

Added support for passing arbitrary values through to the auth URL endpoint.

Added acceptance test scenarios.

### DIFF
--- a/api/teams_test.go
+++ b/api/teams_test.go
@@ -737,9 +737,19 @@ var _ = Describe("Teams API", func() {
 							team.BasicAuth = &basicAuth
 						})
 
-						It("updates the basic auth for that team", func() {
+						It("creates a team with basic auth credentials", func() {
 							Expect(response.StatusCode).To(Equal(http.StatusCreated))
 							Expect(teamServerDB.CreateTeamCallCount()).To(Equal(1))
+
+							createdTeam := teamServerDB.CreateTeamArgsForCall(0)
+							Expect(createdTeam).To(Not(BeNil()))
+
+							Expect(createdTeam.BasicAuth).To(Not(BeNil()))
+							Expect(createdTeam.BasicAuth.BasicAuthUsername).To(Equal("Dean Venture"))
+							Expect(createdTeam.BasicAuth.BasicAuthPassword).To(Equal("Giant Boy Detective"))
+							Expect(createdTeam.GitHubAuth).To(BeNil())
+							Expect(createdTeam.UAAAuth).To(BeNil())
+							Expect(createdTeam.GenericOAuth).To(BeNil())
 						})
 					})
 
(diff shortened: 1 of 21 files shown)


poida:
For a better example of what `--generic-oauth-auth-param` could be used for, GitHub lets you disable signups from the oauth authentication page by passing `allow_signup=false` - https://developer.github.com/v3/oauth/.

vito:
Looking good, I think there are some missing tests though. e.g. here: https://github.com/concourse/atc/blob/master/db/team_db_test.go#L303

May want to just grep for `uaa` and see where it had tests added - will probably be around the same amount.

poida:
@vito true that! I'll take a look and add tests to match.

api/teams_test.go
```diff
@@ -636,6 +733,17 @@ var _ = Describe("Teams API", func() {
...
 					})
+
+					Context("when passed Generic OAuth credentials", func() {
+						BeforeEach(func() {
+							team.GenericOAuth = &genericOAuth
+						})
+
+						It("updates the Generic OAuth auth for that team", func() {
+							Expect(response.StatusCode).To(Equal(http.StatusCreated))
+							Expect(teamServerDB.CreateTeamCallCount()).To(Equal(1))
```
vito: Hmm. Is this actually covering the right thing? It doesn't make any assertions about the auth bit. AFAICT that method (`CreateTeam`) is called regardless; `UpdateGenericOAuth` is what to look for. Looks like the other tests might be similarly broken.

vito:
cool, new tests look pretty good; left one inline comment though.

BTW for a full end-to-end thing you'll want to update fly's [`set-team`](https://github.com/concourse/fly/blob/master/commands/set_team.go) command as well.

api/teams_test.go
poida: @vito the context here is that the team doesn't exist (line 652). Because the team doesn't exist, `CreateTeam` is called. The `Update*Auth` methods are not called in this case, setting the the team server is relying on the db to set it up correctly when it creates it (in set.go).

This is in contrast to the update tests below that do the check as you described:
```
It("updates the Generic OAuth auth for that team", func() {
  Expect(response.StatusCode).To(Equal(http.StatusCreated))
  Expect(teamServerDB.CreateTeamCallCount()).To(Equal(1))
  Expect(teamDB.UpdateGenericOAuthCallCount()).To(Equal(1))
})
```

So it looks like its just asserting the dependency is being called. How about we rename the test so the intent is clearer? Something like:
```
Context("when passed Generic OAuth credentials", func() {
It("creates the team", func() {
```

poida:
@vito I've tweaked those couple of API tests you mentioned in your comment. What do you think?

vito:
Those changes look great, thanks!

Mind rebasing? We've done a bit of work in the auth area, looks like there are conflicts now.

As a bonus, now we'll get to see our shiny new Pull Request pipeline in action:

https://ci.concourse.ci/teams/main/pipelines/prs

poida:
@vito I've rebased. Please review when you are free.

vito:
thanks for this! should be in 2.0

PR #104, opened 2016-08-23
add riemann service prefix

mastertinner:
This PR is a prerequisite for https://github.com/concourse/concourse/pull/602.


vito:
Mind explaining a bit what the use case for this is? Looks good though (aside from the fact that it no longer merges cleanly - my fault for not getting to it sooner).

fdhex:
As a Riemann server can accept many sources, an optional prefix allows for easier distinction between metrics coming from Concourse vs. metrics coming from other inputs, e.g. using a generic (where (service "#concourse/*") ( ... ))

vito:
Sounds reasonable - thanks!

mastertinner:
Thanks, @vito!

PR #112, opened 2016-09-13
Implemention of scope verification for oauth authorization

LinuxBozo:
This adds an optional way to require a particular scope for authorization of a user when using the generic oauth provider.

### DIFF
--- a/auth/genericoauth/provider.go
+++ b/auth/genericoauth/provider.go
@@ -28,8 +28,15 @@ func NewProvider(
 		endpoint.TokenURL = genericOAuth.TokenURL
 	}
 
+	var oauthVerifier verifier.Verifier
+	if genericOAuth.Scope != "" {
+		oauthVerifier = ScopeVerifier{scope: genericOAuth.Scope}
+	} else {
+		oauthVerifier = NoopVerifier{}
+	}
+
 	return Provider{
-		Verifier: NoopVerifier{},
+		Verifier: oauthVerifier,
 		Config: ConfigOverride{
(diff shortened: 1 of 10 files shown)


auth/genericoauth/provider.go
```diff
@@ -28,8 +28,15 @@ func NewProvider(
 		endpoint.TokenURL = genericOAuth.TokenURL
 	}
 
+	var oauthVerifier verifier.Verifier
+	if genericOAuth.Scope != "" {
+		oauthVerifier = ScopeVerifier{scope: genericOAuth.Scope}
```
vito: This should probably use `NewScopeVerifier` - filling in the private field seems a little sketchy.

Or just make the field public; the constructor isn't doing much?

vito:
Nice!

Would it make sense to configure multiple scopes? I'm not sure if that's a common convention, and if it'd be an "AND" or an "OR", but I thought I'd ask.

Left one inline comment. The rest looks good.

auth/genericoauth/provider.go
LinuxBozo: Agreed. Updated.

LinuxBozo:
I considered multiple scopes at first, but then ran into your same thought: and/or. Figured it was best just to get something as far as verification in the codebase and let someone else worry about that if they needed it. My particular use case is served by a single scope assignment of `concourse.admin` to anyone authorized for access to the `main` team.

vito:
Looks good - thanks!

PR #115, opened 2016-09-19
Treat workers with tag "default" as if they had no tags.

SHyx0rmZ:
Sometimes we want to tag our workers, but still want them to be part of the
default pool. In this case the tags can be considered as merely conveying
additional information.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -684,7 +684,7 @@ func (worker *gardenWorker) Uptime() time.Duration {
 
 func (worker *gardenWorker) tagsMatch(tags []string) bool {
 	if len(worker.tags) > 0 && len(tags) == 0 {
-		return false
+		tags = []string{"default"}
 	}
 
 insert_coin:
--- a/worker/worker_test.go
+++ b/worker/worker_test.go
@@ -2033,11 +2033,26 @@ var _ = Describe("Worker", func() {
 					spec.Tags = nil
 				})
 
-				It("returns ErrIncompatiblePlatform", func() {
-					Expect(satisfyingErr).To(Equal(ErrMismatchedTags))
+				Context("when the worker does not have the \"default\" tag", func() {
+					It("returns ErrIncompatiblePlatform", func() {
+						Expect(satisfyingErr).To(Equal(ErrMismatchedTags))
+					})
+				})
+
+				Context("when the worker has the \"default\" tag", func() {
+					BeforeEach(func() {
+						tags = []string{"some", "tags", "default"}
+					})
+
+					It("returns the worker", func() {
+						Expect(satisfyingWorker).To(Equal(gardenWorker))
+					})
+
+					It("returns no error", func() {
+						Expect(satisfyingErr).NotTo(HaveOccurred())
+					})
 				})
 			})
-
 			Context("when the worker has no tags", func() {
 				BeforeEach(func() {
 					tags = []string{}


vito:
Hrm, I think this should be done differently, but I'm not sure how. It just feels weird to have tags default to something. Maybe an explicit property saying "keep me in the default pool"? Or replace tags with something more expressive, i.e. individual properties which can say whether they're required to be specified for placement? I'm not a huge fan that we picked "tags" as the name since it's so generic.

SHyx0rmZ:
Ok, let me explain my current use-case, so my (non-)ideas may make a bit more sense. I currently have a pipeline that builds Debian packages for Concourse and now want to automatically upgrade all instances once a new package has been uploaded. For that to work properly, I figured I needed to have some mechanism in place for a worker to not upgrade itself (because it would restart the running process and abort the upgrade build).

I initially wanted to have each worker tagged with a `I-can't-deploy-<workers-hostname>` tag, but realized that tags can't be used for exclusion, so decided that I'd tag *all* instances besides the worker with `I-can-deploy-<workers-hostname>`. Then I would only need two or more workers (which I have).

When testing, I then noticed that having those tags removed my workers from the default pool, which is not what I wanted, because all my other pipelines need to run, too. At this point I figured I could submit a pull request that would do _something_ about it. My initial idea was to propose a breaking change, so workers would not drop out of the default pool when being tagged.

This, of course, is a bad idea. I then decided the next-best thing would be to have an option to keep the worker in the default pool. I tried to give the worker the `default` tag, but only later realized that that would not work. So I modified the code to enable exactly this behavior and called it a day.

From what you are suggesting, I therefore most like the idea for an explicit property saying to keep the worker in the default pool. To be fair, I don't quite understand what you are suggesting with the individual properties, but there is another crazy thought I had, which might touch upon the same idea:

Instead of just specifying a tag, a step could have a section called `worker`, which could then have many children. One of the possible children would be `tag`, if one wanted to keep the tagging functionality. Other possible children would include properties concerning the worker, maybe fed from the metrics system. It would also be nice to have operators like `not`, `and` and `or`.

I don't think the above would be a good idea. It would introduce a breaking change and reduce maintainability, because we would need to give some sort of guarantee for a stable set of properties.

All in all, a new flag to keep the worker in the default pool sounds like the best idea to me. Also, I don't think that "tags" is a bad name 😄

pivotal-issuemaster:
@SHyx0rmZ Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=115) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@SHyx0rmZ Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=115)!

pivotal-issuemaster:
@SHyx0rmZ Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=115)!

pivotal-issuemaster:
@SHyx0rmZ Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=115)!

vito:
Sorry for the spam, some trouble with the Tracker API is bubbling up in weird ways.

As a side note while I'm here, I think it'd be nice if we renamed "tags" to "capabilities" as part of whatever direction we go with here. Today's `--tag a` would become `--require-capability a` or something, with `--capability` for your use case of saying "this worker can do X but can also do everything else just fine".

SHyx0rmZ:
Should the steps in the build plan also be changed to `capabilities`?

jtarchie:
@SHyx0rmZ, please open a new issue discussing this feature. It appears that there is more to discuss beyond this original PR.

We are considering this pull request dead due to inactivity. Thanks!

Issue #86, opened 2016-06-07
As a Concourse user, I would like Concourse to fail validation of my configuration if there are unused resources


mmb:
When I run set-pipeline with a pipeline config that has resources defined that are not used in any jobs I would l like it to fail validation.

In the same vein as golang refusing to compile with unused imports and variables, this can reveal bugs, reduce cruft and make it easier to clean up dependencies of jobs that have been removed.

Willing to send PR.

mmb:
I wrote some code to generate warnings for unused resources, but they show up in fly as deprecation warnings. It could be easily changed to generate errors.

@vito Is there any version of this that you would accept a PR for?

vito:
Yeah, I like the idea of having that fail validation. Thanks for working on this!

vito:
closing this since we merged the PR

PR #116, opened 2016-09-21
Unused resources fail config validation

mmb:
Finishes #86


mmb:
Thanks for review @joshzarrabi and @vito. I changed this to use JobInputs and JobOutputs.

vito:
looks great, thanks!

PR #117, opened 2016-09-21
Expose new environment variable BUILD_TEAM_NAME

SHyx0rmZ:
This exposes the new environment variable `BUILD_TEAM_NAME`, so you can people where to find their logs.


vito:
thanks!

PR #125, opened 2016-10-15
Attach callbacks to jobs as well as steps.

jmcarp:
Submitting for feedback before writing tests--lmk if the approach makes sense.

[Resolves https://github.com/concourse/concourse/issues/166]

### DIFF
--- a/config.go
+++ b/config.go
@@ -87,15 +87,10 @@ func (types ResourceTypes) Lookup(name string) (ResourceType, bool) {
 	return ResourceType{}, false
 }
 
-type CallbackConfig struct {
-	// used by any step or plan to run something when the step reports a failure
-	Failure *PlanConfig `yaml:"on_failure,omitempty" json:"on_failure,omitempty" mapstructure:"on_failure"`
-
-	// used on any step or plan to always execute regardless of the step's completed state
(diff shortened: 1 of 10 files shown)


vito:
Sweet!

This seems worth of additional testing in [Testflight](https://github.com/concourse/testflight) - probably mirroring the behavior of https://github.com/concourse/testflight/blob/master/pipelines/hooks_test.go

vito:
GitHub's UI is confusing me here. It looks like there's a commit (6d84027) that adds unit tests, but those changes don't show up when looking at the "Files changed" tab.

vito:
Is there anything weird that happens in this case?

```sh
jobs:
- name: double-hooked
  on_failure: ...
  plan:
  - task: x
    on_failure: ...
```

I'm wondering if one of them clobbers the other. Haven't examined the code thoroughly but I could see that happening with certain approaches.

jmcarp:
Sorry, forgot to check in the test somehow--just updated. The test should cover the case that you described. Same goes for the PR against testflight, which is passing locally.

vito:
Ah, so it does. Thanks!

jraqula:
thanks @jmcarp for this contribution! i've merged it into the pipeline and will close this PR.

PR #128, opened 2016-10-27
When many groups are created the page rendering is strange

srbry:
### What is it?

When displaying a large number of groups with names that combine to be wider than the page, group names are split over multiple lines and do not render right at all.

This PR is to make the navbar horizontally scrollable while pinning the user menu to the top right at all times.

### DIFF
--- a/web/assets/css/layout.less
+++ b/web/assets/css/layout.less
@@ -31,7 +31,8 @@ html, body, #elm-app-embed, #subpage, .content-frame {
       flex-shrink: 0;
       overflow-x: scroll;
       overflow-y: hidden;
-      width: 85%;
+      position: fixed;
+      right: 120px;
 
       li {
         flex-shrink: 0;
--- a/web/assets/css/layout.less
+++ b/web/assets/css/layout.less
@@ -40,6 +40,19 @@ html, body, #elm-app-embed, #subpage, .content-frame {
       }
     }
 
+    .groups::-webkit-scrollbar-track {
+      background-color: #1A252F;
+    }
+
+    .groups::-webkit-scrollbar {
+      height: 4px;
+      background-color: #1A252F;
+    }
+
+    .groups::-webkit-scrollbar-thumb {
+      background-color: #E6E7E8;
+    }
+
     .nav-right {
       flex-shrink: 1;
       position: fixed;
--- a/web/assets/css/layout.less
+++ b/web/assets/css/layout.less
@@ -38,19 +38,19 @@ html, body, #elm-app-embed, #subpage, .content-frame {
       li {
         flex-shrink: 0;
       }
-    }
 
-    .groups::-webkit-scrollbar-track {
-      background-color: #1A252F;
-    }
+      ::-webkit-scrollbar-track {
+        background-color: #1A252F;
+      }
 
-    .groups::-webkit-scrollbar {
-      height: 4px;
-      background-color: #1A252F;
-    }
+      ::-webkit-scrollbar {
+        height: 4px;
+        background-color: #1A252F;
+      }
 
-    .groups::-webkit-scrollbar-thumb {
-      background-color: #E6E7E8;
+      ::-webkit-scrollbar-thumb {
+        background-color: #E6E7E8;
+      }
     }
 
     .nav-right {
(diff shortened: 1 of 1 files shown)


srbry:
Just noticed that these changes only render nicely if a browser can detect a mouse has side scrolling. I.E the mac magic mouse.

Otherwise a scrollbar is permanently visible.

srbry:
Scrollbar should now always be visible when there is overflowing content and actually fit in with the concourse design.

srbry:
Tests have failed trying to fetch package: `code.cloudfoundry.org/garden/client/connection/fakes` think that package should be: `code.cloudfoundry.org/garden/client/connection/connectionfakes` as per the garden repos. This isn't anything we have changed for this PR, is something in the ci broken?

jimconner:
Awesome - this makes things look so much better and clears up the terrible mess at the top of the page when you have a lot of groups. nice work @srbry

srbry:
### Before:
screen shot 2016-11-16 at 14 20 14
### After:
screen shot 2016-11-16 at 14 20 40

rainmaker:
Hi all, 

My team is also experiencing this unfortunate issue, mostly when we try to view our build pages. Pretty frustrating to try and read the build info, or to retrigger it. Wanted to +1 this issue as a whole.


screen shot 2016-11-17 at 2 18 39 pm


I'm glad to see that there's a discussion and an implemented solution, but not sure whether horizontal scroll would be the best option. In my experience, horizontal scroll isn't very widely used and can be tricky depending on your mouse. Our team would be fine if the groups were given n extra rows to be listed fully, and the rest of the build info (the yellow bar and below) were to start after that. However, we'd be happy with whatever solution makes sense to the community and Concourse team. We'll be following to see what comes out of this.

Thanks!

Raina
PCF RelEng

srbry:
Hi @rainmaker I do agree completely, horizontal scrolling was the easiest thing for us to put in with just changing some css. We made sure the scroll bar shows properly when it needs too so you can click and drag it so it works with all mice, although it works far nicer with a mouse that can do horizontal scrolling properly, if you want to play with the format you can use a chrome plugin called `stylish` and apply this css override:
```
.content-frame .top-bar .groups {
    flex-grow: 0;
    display: flex;
    flex-direction: row;
    flex-shrink: 0;
    overflow-x: auto;
    overflow-y: hidden;
    position: fixed;
    right: 120px;
    left: 0;
}

.content-frame .top-bar .groups li {
    flex-shrink: 0;
}
.content-frame .top-bar .nav-right {
    flex-shrink: 1;
    position: fixed;
    right: 0;
}
.content-frame .top-bar .groups::-webkit-scrollbar-track {
    background-color: #1A252F;
}
.content-frame .top-bar .groups::-webkit-scrollbar {
    height: 4px;
    background-color: #1A252F;
}
.content-frame .top-bar .groups::-webkit-scrollbar-thumb {
    background-color: #E6E7E8;
}
```

thelangley:
+1. good fix thanks srbry

pivotal-issuemaster:
@srbry Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=128) to manually synchronize the status of this Pull Request.

vito:
@srbry This looks good! Sorry for the inactivity, we kinda got ahead of ourselves and had UI changes planned for this area so we hesitated on pulling this in, but it probably makes sense to at least make the current UX better immediately first.

Tried out the CSS changes myself and confirmed that it only affects the given use case (i.e. doesn't introduce any UI changes until the scrollbar needs to appear), so this looks good to me. It would be nice to have vertical scrolling translate to horizontal scrolling as it does on the build page (for the builds under the job header), but that's surprisingly annoying to implement and involves JS, so I don't mind this as-is as it's strictly an improvement over today's UX.

Our CLA bot isn't happy though. Could you look at signing it?

chendrix:
@srbry can you please look at the CLA if you'd like this merged?

sambryant4:
This has been re-raised under: https://github.com/concourse/atc/pull/174, the CLA has now been signed

PR #137, opened 2016-11-17
Use team.slug for GitHub auth teams matching

stepanstipl:
Currently Concourse is using `team.name` returned by GitHub's API for matching teams, this is causing issues when team name contain special characters, as these are encoded.

For example if you want to auth against team named `Monkeys & Bananas` under organization `myorg`, team has to be passed to Concourse as `myorg/Monkeys \u0026 Bananas`, else it's not matched.

GitHub teams API returns also `team.slug`, which is machine friendly version (also used in GitHub's urls) and would look like `monkeys-bananas` (see https://developer.github.com/v3/orgs/teams/#list-user-teams). I would suggest using this preferably to the name.

*Please note that this is potentially breaking change for current configuration.*

### DIFF
--- a/auth/github/client.go
+++ b/auth/github/client.go
@@ -62,7 +62,9 @@ func (c *client) Teams(httpClient *http.Client) (OrganizationTeams, error) {
 				organizationTeams[organizationName] = []string{}
 			}
 
+			// We add both forms (slug and name) of team
 			organizationTeams[organizationName] = append(organizationTeams[organizationName], *team.Name)
+			organizationTeams[organizationName] = append(organizationTeams[organizationName], *team.Slug)
 		}
 
 		nextPage = resp.NextPage
--- a/auth/github/client_test.go
+++ b/auth/github/client_test.go
@@ -176,8 +176,8 @@ var _ = Describe("Client", func() {
 				teams, err := client.Teams(proxiedClient)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(teams).To(HaveLen(2))
-				Expect(teams["org-1"]).To(ConsistOf([]string{"Team 1", "Team 2"}))
-				Expect(teams["org-2"]).To(ConsistOf([]string{"Team 3"}))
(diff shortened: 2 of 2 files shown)


vito:
Yeah let's not break existing configurations. Could it support both?

stepanstipl:
Sure can do that...

stepanstipl:
@vito This should now add team in both forms, so we stay backward compatible (I would still eventually get rid of the name form potentially containing special characters, but that's just personal preferencs :))

vito:
Mind rebasing this to kick CI again? I fixed the error it's reporting (it was unrelated). There were also no tests updated for this, so I suspect one is broken or at least needs to be updated/added.

stepanstipl:
np, I'll rebase and try to add some test for this

stepanstipl:
@vito I did rebase on master, but the PR tests seems to fail still with:
```
package code.cloudfoundry.org/garden/client/connection/fakes: cannot find package "code.cloudfoundry.org/garden/client/connection/fakes" in any of:
	/usr/local/go/src/code.cloudfoundry.org/garden/client/connection/fakes (from $GOROOT)
	/tmp/build/0df9eea0/gopath/src/code.cloudfoundry.org/garden/client/connection/fakes (from $GOPATH)
```
am I missing something?

Updated the tests for this as well.

pivotal-issuemaster:
@stepanstipl Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=137) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@stepanstipl Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=137)!

stepanstipl:
Rebased on latest master, the tests are green now.

vito:
thanks!

PR #140, opened 2016-11-26
Add X-Accel-Buffering header to SSE endpoints

### LINKED ISSUE #139: Add X-Accel-Buffering header to SSE endpoints
When deploying concourse behind nginx most browsers have a difficult time with the SSE event streams because of response buffering. Testing on macOS, Chrome and Chrome Canary fail with the error `net::ERR_INCOMPLETE_CHUNKED_ENCODING` while Safari has no problem. On Windows, Edge, IE11, and Chrome all failed while Firefox has no problem.

When looking for a solution I found this [server fault answer](http://serverfault.com/questions/801628/for-server-sent-events-sse-what-nginx-proxy-configuration-is-appropriate#801629) that goes into detail about how to handle SSE behind nginx.

There is a header you can add to tell reverse proxies to not buffer the response: `X-Accel-Buffering: no`. This should be added to all SSE endpoints.

[Nginx docs on `X-Accel-Buffering`](https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/#x-accel-buffering)

jasonkeene:
Rational for this is covered in #139.

### DIFF
--- a/api/buildserver/eventhandler.go
+++ b/api/buildserver/eventhandler.go
@@ -35,6 +35,7 @@ func NewEventHandler(logger lager.Logger, build db.Build) http.Handler {
 
 		w.Header().Add("Content-Type", "text/event-stream; charset=utf-8")
 		w.Header().Add("Cache-Control", "no-cache, no-store, must-revalidate")
+		w.Header().Add("X-Accel-Buffering", "no")
 		w.Header().Add(ProtocolVersionHeader, CurrentProtocolVersion)
 
 		writer := eventWriter{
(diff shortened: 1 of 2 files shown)


vito:
great find! looking forward to merging this.

would you mind rebasing, just to get the CI build going? there was an import path issue which I've just fixed.

thanks!

jasonkeene:
Rebased

jasonkeene:
Looks like the acceptance suite failed for an unrelated reason. Prior builds are failing as well. I ran the suite locally and it took forever but passed.

vito:
We'll look into that today - looks like Baggageclaim needs to be updated.

vito:
@jasonkeene I'll merge this since the failures are definitely unrelated. Thanks!

PR #149, opened 2016-12-28
Validate config

jmcarp:
Updated according to review from @vito, lmk if this needs more updates!


vito:
Great, thanks a bunch for doing this!

PR #163, opened 2017-04-11
Feature webhook

mainephd:
Addresses issue [#331](https://github.com/concourse/concourse/issues/331) 

The resource DSL has been updated to include a `webhook_token` attribute that allows the end user to configure an arbitrary string to be used to trigger a resource to perform a check.

The webhook would be `http://{yourserver}/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name/check/webhook?access_token=somevalue`

There is also a convenience method to generate a token value
`http://{yourserver}/api/v1/teams/:team_name/auth/access_token`

A separate pull request would need to be submitted to pull in the dependency on https://github.com/satori/go.uuid


pivotal-issuemaster:
@mainephd Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=163)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=163) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mainephd Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=163)!

api/resourceserver/check_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+func (s *Server) CheckResourceWebHook(pipelineDB db.PipelineDB, dbPipeline dbng.Pipeline) http.Handler {
+	logger := s.logger.Session("check-resource-webhook")
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resourceName := rata.Param(r, "resource_name")
+		accessToken := r.URL.Query().Get("access_token")
+
+		if accessToken == "" {
+			logger.Info("malformed-request", lager.Data{"error": "missing access_token"})
+			w.WriteHeader(http.StatusBadRequest)
```
vito: Should this be a 401?

api/resourceserver/check_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+// CheckResourceWebHook defines a handler for process a check resource request via an access token.
+func (s *Server) CheckResourceWebHook(pipelineDB db.PipelineDB, dbPipeline dbng.Pipeline) http.Handler {
+	logger := s.logger.Session("check-resource-webhook")
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resourceName := rata.Param(r, "resource_name")
+		accessToken := r.URL.Query().Get("access_token")
+
+		if accessToken == "" {
+			logger.Info("malformed-request", lager.Data{"error": "missing access_token"})
```
vito: This can probably just say `no-access-token` (or `no-password`, given my larger comment) and not log the error bit.

api/resourceserver/check_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+
+		if err != nil {
+			logger.Info("database-error", lager.Data{"error": err})
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		token := pipelineResource.Config.WebhookToken
+		if token != accessToken {
+			logger.Info("invalid-token-error", lager.Data{"error": fmt.Sprintf("Actual token %s does not match expected token %s", accessToken, token)})
```
vito: No need for the `error` bit here; I don't think we should log either token.

api/resourceserver/check_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+		scanner := s.scannerFactory.NewResourceScanner(pipelineDB, dbPipeline)
+		err = scanner.ScanFromVersion(logger, resourceName, nil)
+		switch scanErr := err.(type) {
+		case resource.ErrResourceScriptFailed:
+			checkResponseBody := atc.CheckResponseBody{
+				ExitStatus: scanErr.ExitStatus,
+				Stderr:     scanErr.Stderr,
+			}
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusBadRequest)
```
vito: This shouldn't be a 400 error. Maybe 500? But maybe this is existing bad UX with the checking endpoint; I'm guessing this was copypasta.

We should also not print the stderr for the webhook case; if someone manages to acquire the password, this could lead to further information leakage.

How about just check for an error and log it? Same with the `case error` branch. Guess this case could just be removed at that point.

wrappa/api_auth_wrappa.go
```diff
@@ -140,12 +142,17 @@ func (wrappa *APIAuthWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
...
 		}
 
-		if name == atc.GetAuthToken {
+		if name == atc.GetAuthToken || name == atc.GetAccessToken {
 			newHandler = auth.WrapHandler(newHandler, wrappa.getTokenValidator, wrappa.userContextReader)
 		} else {
 			newHandler = auth.WrapHandler(newHandler, wrappa.authValidator, wrappa.userContextReader)
 		}
-		wrapped[name] = auth.CSRFValidationHandler(newHandler, rejector, wrappa.userContextReader)
+		if name != atc.CheckResourceWebHook {
```
vito: Why's this needed?

billimek: We were expecting this question!  @mainephd can probably provide more context, but:

`CSRFValidationHandler` expects the client to be authenticated for all GET API calls (e.g. DownloadCLI, ListAllPipelines, etc).  Issue here is that the webhook is a POST, and won't get the exception inside `CSRFValidationHandler`, hence this check.

api/resourceserver/check_webhook.go
mainephd: If the `basic-auth` approach is taken would the built in `basic_auth_validator` need to be used or setup as some middleware for this endpoint? The logger statement is easy enough to change so i'll make the change.

api/resourceserver/check_webhook.go
mainephd: i waffled on whether to send a 400 vs a 401. It technically is a bad request because all of the required attributes aren't present but based on your equating the token to a password then 401 make sense. So I can go either way.

api/resourceserver/check_webhook.go
mainephd: Agreed.

api/resourceserver/check_webhook.go
mainephd: This was indeed copypasta. :) So if an error occurs (essentially the all of the `switch` statements) then just return a 500 error?

wrappa/api_auth_wrappa.go
mainephd: Without this special case the webhook caller would have to include the `X-Csrf-Token` header on each request and shouldn't be available for the caller to access otherwise the CSRF protection becomes useless. So thus we added the special case. We wanted to do it more generically but that would have been a bigger change.

api/resourceserver/check_webhook.go
mainephd: And you are welcome. It was cool to finally dig into the code that has made our lives ALOT easier to manage in terms of development.

wrappa/api_auth_wrappa.go
vito: Hmm I don't think this will be required for the actual use-case though. The `X-Csrf-Token` validation is only checked if the user has the `ATC-Authorization` cookie present. Requests made to this endpoint will have neither `Authorization` nor `ATC-Authorization`, as the endpoint has its own special auth rules (the query param).

api/resourceserver/check_webhook.go
```diff
@@ -0,0 +1,58 @@
...
+
+		token := pipelineResource.Config.WebhookToken
+		if token != accessToken {
+			logger.Info("invalid-token", lager.Data{"error": fmt.Sprintf("invalid token for webhook %s", accessToken)})
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		scanner := s.scannerFactory.NewResourceScanner(pipelineDB, dbPipeline)
+		err = scanner.ScanFromVersion(logger, resourceName, nil)
```
vito: This should actually check from the current version, as in https://github.com/concourse/atc/blob/704dd8511bb331d6cc2793cda25252a1cd51a5a3/api/resourceserver/check.go#L29-L45

The reason for this is that a hook may be triggered by someone pushing 100 commits. We want to find all the ones since the current version, not just the last of the 100 commits.

api/resourceserver/check_webhook.go
vito: I wouldn't lean on the existing BasicAuthValidator as that is specifically for team-configured basic auth. This is sort of its own way of checking auth, particular to this endpoint, so it could just interpret the `Authorization` header itself, with it having been ignored by all the other middleware.

Maybe it could just statically check for `webhook` as the user and the configured password as the password?

auth/cookie_set_handler.go
```diff
@@ -14,6 +14,7 @@ type CookieSetHandler struct {
 
 func (handler CookieSetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	cookie, err := r.Cookie(AuthCookieName)
+
```
vito: Don't need this change; mind rebasing it out just to reduce files touched?

task.go
```diff
@@ -330,7 +330,7 @@ func (config TaskConfig) validateInputContainsNames() []string {
 type TaskRunConfig struct {
 	Path string   `json:"path" yaml:"path"`
 	Args []string `json:"args,omitempty" yaml:"args"`
-	Dir  string   `json:"dir",omitempty" yaml:"dir"`
```
vito: Nice catch, though did you see this breaking things? Should this be its own release note?

web/bindata.go
```diff
@@ -112,7 +112,7 @@ func indexHtml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "index.html", size: 2615, mode: os.FileMode(420), modTime: time.Unix(1491867544, 0)}
```
vito: These changes probably shouldn't be here?

auth/cookie_set_handler.go
mainephd: Sure

web/bindata.go
mainephd: I am not sure how this line was removed. I'll have to do some digging to see how it got removed.

task.go
mainephd: When i ran the unit test locally it did not break things. It probably should be noted in a release cause it could be broken....

wrappa/api_auth_wrappa.go
mainephd: So because of this line https://github.com/concourse/atc/blob/master/auth/cookie_set_handler.go#L16 it would only check that the `X-Csrf-Token` is specified. I have only tested locally in `no-auth` mode.

mainephd:
The acceptance test seems to be behave more consistently locally when i change the default timeout and polling configurations of the `Eventually` assertions from

```
SetDefaultEventuallyTimeout(10 * time.Second)
SetDefaultEventuallyPollingInterval(100 * time.Millisecond)
```

 to

```
 SetDefaultEventuallyTimeout(20 * time.Second)
 SetDefaultEventuallyPollingInterval(1 * time.Second)
```

task.go
vito: I'll track this as a separate bug and see if it actually affected anything. Thanks!

api/resourceserver/check_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/tedsuo/rata"
+)
+
+// CheckResourceWebHook defines a handler for process a check resource request via an access token.
+func (s *Server) CheckResourceWebHook(pipelineDB db.PipelineDB, dbPipeline dbng.Pipeline) http.Handler {
+	logger := s.logger.Session("check-resource-webhook")
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resourceName := rata.Param(r, "resource_name")
+		accessToken := r.URL.Query().Get("access_token")
```
vito: Can we rename this to `webhook_token`? It has specific handling for this endpoint, whereas `access_token` sounds like a generic API for passing a token to all handlers.

mainephd: That is easy enough.

task.go
mainephd: So I am clear. Do you want me to revert my change or leave it in tact?

vito: You can leave it - the separate bug is in our Tracker just so I can do explicit acceptance on it.

vito:
Awesome, thanks! 🎊

vito:
The CI failure is in `acceptance/` - not your fault. I've logged a bug to have us clean that up, it's really annoying for PRs. Merging!

trivigy:
What is the right way to deal with `check_every` when setting the `webhook_token`?

PR #164, opened 2017-04-21
[WIP] Authorities for generic oauth client

mvdkleijn:
I'm entirely new to Go so there's probably lot's more needed that I'm not aware of, but this seemed like a decent enough first start. Please note that I'd be happy to get support from others on this PR! :smile: 

This basically attempts to add basic initial support for a `--generic-oauth-authority <authority name>` argument to fly's `set-team` command.

Related to my feature request: concourse/concourse#1026

### DIFF
--- a/auth/genericoauth/authority_verifier.go
+++ b/auth/genericoauth/authority_verifier.go
@@ -59,8 +59,8 @@ func (verifier AuthorityVerifier) Verify(logger lager.Logger, httpClient *http.C
 		return false, errors.New("user has no assigned authorities in access token")
 	}
 
-	for _, userScope := range oauthToken. {
-		if userScope == verifier.authority {
+	for _, userAuthority := range oauthToken.Authorities {
+		if userAuthority == verifier.authority {
 			return true, nil
 		}
 	}
--- a/auth/genericoauth/authority_verifier.go
+++ b/auth/genericoauth/authority_verifier.go
@@ -1,0 +1,74 @@
+package genericoauth
+
+import (
+	"encoding/json"
+	"errors"
+	"net/http"
+	"strings"
+
+	"golang.org/x/oauth2"
+
+	"code.cloudfoundry.org/lager"
+	"github.com/concourse/atc/auth/verifier"
+)
+
+type AuthorityVerifier struct {
+	authority string
+}
+
+func NewAuthorityVerifier(
+	authority string,
+) verifier.Verifier {
+	return AuthorityVerifier{
+		authority: authority,
+	}
+}
+
+type GenericOAuthToken struct {
+	Authorities []string `json:"authorities"`
+}
+
+func (verifier AuthorityVerifier) Verify(logger lager.Logger, httpClient *http.Client) (bool, error) {
+	oauth2Transport, ok := httpClient.Transport.(*oauth2.Transport)
+	if !ok {
+		return false, errors.New("httpClient transport must be of type oauth2.Transport")
+	}
+
+	token, err := oauth2Transport.Source.Token()
+	if err != nil {
+		return false, err
+	}
+
+	tokenParts := strings.Split(token.AccessToken, ".")
+	if len(tokenParts) < 2 {
+		return false, errors.New("access token contains an invalid number of segments")
+	}
+
+	decodedClaims, err := jwt.DecodeSegment(tokenParts[1])
+	if err != nil {
+		return false, err
+	}
+
+	var oauthToken GenericOAuthToken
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@mvdkleijn Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=164)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=164) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mvdkleijn Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=164)!

auth/genericoauth/provider.go
```diff
@@ -141,9 +142,12 @@ func (GenericTeamProvider) ProviderConstructor(
...
 
 	var oauthVerifier verifier.Verifier
-	if genericOAuth.Scope != "" {
+	if genericOAuth.Scope != "" && genericOAuth.Authority = "" {
 		oauthVerifier = NewScopeVerifier(genericOAuth.Scope)
-	} else {
+	} else if genericOAuth.Scope = "" && genericOAuth.Authority != "" {
+		oauthVerifier = NewAuthorityVerifier(genericOAuth.Authority)
+	}
+	else {
```
vito: Not blocking merge on this, but `go fmt` would clean this up (may want to look for editor integration too - a lot of them can do it automatically on save).

auth/genericoauth/provider.go
```diff
@@ -141,9 +142,12 @@ func (GenericTeamProvider) ProviderConstructor(
 	}
 
 	var oauthVerifier verifier.Verifier
-	if genericOAuth.Scope != "" {
+	if genericOAuth.Scope != "" && genericOAuth.Authority = "" {
```
vito: This will result in configuring no verification at all if both are configured, which seems risky or at least counterintuitive.

Does it make sense to check for both a scope and an authority? Or should that be a validation error? Or should one take precedence?

auth/genericoauth/authority_verifier.go
```diff
@@ -0,0 +1,74 @@
...
+
+func NewAuthorityVerifier(
+	authority string,
+) verifier.Verifier {
+	return AuthorityVerifier{
+		authority: authority,
+	}
+}
+
+type GenericOAuthToken struct {
```
vito: Don't think this'll work as the same type name is already used here within the same package: https://github.com/concourse/atc/blob/ddbaaf7f2f30606dbc3db950c07f24a30d8f7d32/auth/genericoauth/scope_verifier.go#L28

Also, there's a lot of duplication between this verifier and the scope verifier - could this be a refactoring opportunity?

vito:
You may want to consult [CONTRIBUTING.md](https://github.com/concourse/atc/blob/master/CONTRIBUTING.md) by the way. Not sure if you're set up for running tests, since I'm pretty sure at least one of those is a straight up compilation error. :P

auth/genericoauth/provider.go
```diff
@@ -141,9 +142,12 @@ func (GenericTeamProvider) ProviderConstructor(
...
 		oauthVerifier = NewScopeVerifier(genericOAuth.Scope)
-	} else {
+	} else if genericOAuth.Scope = "" && genericOAuth.Authority != "" {
```
vito: That should be a `==`.

auth/genericoauth/provider.go
mvdkleijn: @vito Sorry for the late response and many thanks for the input!

Agreed, it does make it counter intuitive. I believe the correct behaviour would be to check and validate both if the user has set both.

Since it looks like I can only use one verifier at a time, I guess this would require me to write a more generic verifier that would accept multiple inputs?

auth/genericoauth/provider.go
mvdkleijn: Will take a look. I added Go support to my Atom editor and it should run automatically already.

auth/genericoauth/authority_verifier.go
mvdkleijn: I basically copied the scope verifier so you would be right about the code duplication. It seemed the speediest way to get something working. I can probably create a more generic verifier since the scope and authority checks are basically the same.

jtarchie:
We are considering this pull request dead due to inactivity. Please feel free to reopen a new pull request with the requested changes. Thanks!

PR #169, opened 2017-05-15
Added custom paths to use github personal access token for the authen…

Chumper:
…tication

The following things have been changed:

* Add the possibility for providers to add custom endpoints. The `github` provider adds a `/token` endpoint with `POST`
* Pass the logger to all providers to enable creating custom routes.
* Create a new `TokenHandler` for the `GithubbProvider` that can read personal access tokens from the body of a `POST` requests and checks if the user is allowed to access the given team.
* Do the normal authorization and authentication flow that is used for all other methods. I just had to create a different `http.Client` to use with github.

This is just the first part for the `ATC` to support `Personal Accesss Tokens`. 
After the review and potential merge I would enhance the `fly` commandline tool to retreive a token for the `login` command.

This is my first contribution in go, so please go ahead and point out all mistakes and possible bad practies. This one is a `DRAFT` atm.

@vito FYI

I need to delay the Signment of the CLA until my employee Adobe has granted me the right to do so, so please see this as a point of discussion atm.

Edit: Signing the CLA will take approximately 6 weeks.

### DIFF
--- a/auth/routes/oauth_routes.go
+++ b/auth/routes/oauth_routes.go
@@ -6,10 +6,12 @@ const (
 	OAuthBegin    = "OAuthBegin"
 	OAuthCallback = "OAuthCallback"
 	LogOut        = "LogOut"
+	Token         = "Token"
 )
 
 var OAuthRoutes = rata.Routes{
 	{Path: "/auth/logout", Method: "GET", Name: LogOut},
 	{Path: "/auth/:provider/callback", Method: "GET", Name: OAuthCallback},
 	{Path: "/auth/:provider", Method: "GET", Name: OAuthBegin},
+	{Path: "/auth/:provider/token", Method: "POST", Name: Token},
 }
--- a/auth/genericoauth/provider.go
+++ b/auth/genericoauth/provider.go
@@ -1,23 +1,20 @@
 package genericoauth
 
 import (
+	"code.cloudfoundry.org/lager"
+	"encoding/json"
 	"errors"
 	"fmt"
-	"net/http"
-
-	"code.cloudfoundry.org/lager"
-
-	"encoding/json"
-
 	"github.com/concourse/atc"
 	"github.com/concourse/atc/auth/provider"
 	"github.com/concourse/atc/auth/routes"
 	"github.com/concourse/atc/auth/verifier"
 	"github.com/hashicorp/go-multierror"
-	flags "github.com/jessevdk/go-flags"
+	"github.com/jessevdk/go-flags"
 	"github.com/tedsuo/rata"
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2"
+	"net/http"
 )
 
 const ProviderName = "oauth"
--- a/auth/github/provider.go
+++ b/auth/github/provider.go
@@ -1,24 +1,20 @@
 package github
 
 import (
-	"errors"
-	"net/http"
-
-	"golang.org/x/oauth2"
-	"golang.org/x/oauth2/github"
-
-	"fmt"
-	"strings"
-
 	"encoding/json"
-
+	"errors"
+	"fmt"
 	"github.com/concourse/atc"
 	"github.com/concourse/atc/auth/provider"
 	"github.com/concourse/atc/auth/routes"
 	"github.com/concourse/atc/auth/verifier"
 	"github.com/hashicorp/go-multierror"
-	flags "github.com/jessevdk/go-flags"
+	"github.com/jessevdk/go-flags"
 	"github.com/tedsuo/rata"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+	"net/http"
+	"strings"
 )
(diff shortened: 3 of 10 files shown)


pivotal-issuemaster:
@Chumper Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=169)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=169) to manually synchronize the status of this Pull Request.

mariash:
@Chumper please let us know when your CLA will be singed. Thanks!

Chumper:
I updated the PR to include personal access tokens for gitlab.

I am very sorry about the delay with the CLA... Unfortunately there is nothing I can do except keep on asking when I can sign it... 😞

Chumper:
I have signed the CLA as my manager approved the request, so please review this now

pivotal-issuemaster:
@Chumper Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=169)!

Chumper:
@vito @mariash Can you take a look at the PR since I signed the CLA?

jtarchie:
@chumper, I'll be taking a look at this. @mariash has moved teams within Pivotal. Sorry for the delay, we've had some holidays the past few weeks, so playing catch up.

auth/routes/oauth_routes.go
```diff
@@ -6,10 +6,12 @@ const (
...
 	OAuthCallback = "OAuthCallback"
 	LogOut        = "LogOut"
+	PostCustom    = "PostCustom"
 )
 
 var OAuthRoutes = rata.Routes{
 	{Path: "/auth/logout", Method: "GET", Name: LogOut},
 	{Path: "/auth/:provider/callback", Method: "GET", Name: OAuthCallback},
 	{Path: "/auth/:provider", Method: "GET", Name: OAuthBegin},
+	{Path: "/auth/:provider/", Method: "POST", Name: PostCustom},
```
jtarchie: @Chumper, can you explain why this route needs to be added. This seems that it would affect any oauth provider, not just GIthub.

jtarchie:
@chumper, since this a _big_ change, do you have a recommended way of testing this out? I'd like to see the flow of it at the moment.

auth/routes/oauth_routes.go
Chumper: Yes, that is correct, with that we enable all auth providers to add additional routes that can be addressed as POST.
Currently only the `githubProvider` and the `gitlabProvider` are using that to provide an endpoint to authenticate with a token.
You can see that here: https://github.com/Chumper/atc/blob/4fb3dd3c9beba37e46a32967d7843b649b130077/auth/github/provider.go#L182

If no routes are added, a 404 will be thrown, so the default behaviour is the same.

The need to add routes is that by default only a webflow is allowed, that uses `GET auth/:provider` and the callback.

Chumper: The routes can be found here: https://github.com/Chumper/atc/blob/4fb3dd3c9beba37e46a32967d7843b649b130077/auth/github/routes.go

We add a `/token` endpoint to return a bearer token when a valid personal auth token is provided.

Chumper:
I have added tests that should cover the flow on atc side.
If you created a team and acticated `github` auth then you can call ` POST /auth/github/token` with the token in the body to get a valid bearer token in return.

The next step would be to add a switch to fly to check for a `--token` on `fly login` and if present call the token endpoint on the provider (if the provider supports that).
This would be the point to check: https://github.com/concourse/fly/blob/master/commands/login.go#L180

So instead of waiting for the callback we will retrieve a token actively by calling the token endpoint.

Does that make sense so far?

auth/routes/oauth_routes.go
Chumper: The routes are evaluated here: https://github.com/concourse/atc/pull/169/files#diff-a0a797688867b8d2b038ac457091c196R84
If a provider provides no routes (default `nil`) then the endpoint will not be available

jtarchie:
@chumper, thanks for pairing on your code today. After reviewing everything with another concourse dev, we would feel more comfortable if the github/gitlab routes were more explicit. The `CustomRoutes` introduces complexity and some duplication (see comments on `FindTeam`) that we don't think should exist at the moment.

Are you able to take care of that? We can set up a time and pair with you on it.

Chumper:
We can add the token route explicitly to the existing routes here:
https://github.com/concourse/atc/blob/master/auth/routes/oauth_routes.go

But that would mean, that the token route exists for all providers.
I added the custom handler so that the provider can add additional routes without relying cluttering the global routes file.

If you feel like it should be an explicit route I sure can do that. Is that your suggestion?

Chumper:
I have updated the code to use an explicit route. Please have a look.

PR #173, opened 2017-06-05
Convert type to string so that it can be converted by library

kurtmc:
I saw some messages in the ATC log like this:
```
{
  "timestamp": "1496628557.047648191",
  "source": "atc",
  "message": "atc.build-tracker.track.failed-to-lookup-started-builds",
  "log_level": 2,
  "data": {
    "error": "sql: converting Exec argument #0's type: unsupported type db.BuildStatus, a string",
    "session": "17.4"
  }
}
```

And I saw an example query here https://github.com/concourse/atc/blob/master/db/build.go#L259 where it is converting the type to a string. So I have made the same change to the Abort function, which stops this error from happening.

### DIFF
--- a/db/build.go
+++ b/db/build.go
@@ -365,7 +365,7 @@ func (b *build) Delete() (bool, error) {
 
 func (b *build) Abort() error {
 	_, err := psql.Update("builds").
-		Set("status", BuildStatusAborted).
+		Set("status", string(BuildStatusAborted)).
 		Where(sq.Eq{"id": b.id}).
 		RunWith(b.conn).
 		Exec()


chendrix:
Closed via https://github.com/concourse/concourse/commit/ff830e233afee5f2f2baf2327c0d817924a4c46a

vito:
I'm pretty sure this depends on the version of Go the ATC was built with, and we have other instances of that same thing in other parts of `db`.

What platform was this running on? This was a problem when running the ATC on Windows a while back, but the latest binaries should be OK.

kurtmc:
@vito I built this with go 1.7.5 on Fedora 25. Should I be using go 1.8?

chendrix:
@kurtmc is there a reason you are building the binaries yourself?

And yes, you should be using go1.8

kurtmc:
@chendrix Yes. Well I guess so. I'm using the scripts in https://github.com/concourse/concourse/tree/master/dev

kurtmc:
I'm just building it locally so I can test my changes

chendrix:
@kurtmc Gotcha!, yeah we've moved to using go1.8, I've updated the contributing docs to reflect that.

PR #174, opened 2017-06-20
When many groups are created the page rendering is strange

sambryant4:
### What is it?

When displaying a large number of groups with names that combine to be wider than the page, group names are split over multiple lines and do not render right at all.

This PR is to make the navbar horizontally scrollable while pinning the user menu to the top right at all times.

Re-raising this due to CLA

### DIFF
--- a/web/assets/css/layout.less
+++ b/web/assets/css/layout.less
@@ -31,7 +31,8 @@ html, body, #elm-app-embed, #subpage, .content-frame {
       flex-shrink: 0;
       overflow-x: scroll;
       overflow-y: hidden;
-      width: 85%;
+      position: fixed;
+      right: 120px;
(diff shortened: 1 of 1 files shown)


chendrix:
Validate that the tests work with changes to the ATC

Verify that bindata.go has been updated.

mhuangpivotal:
We are seeing the horizontal scroll bar always displayed blocking the menu (macOS, Chrome).

image

jtarchie:
The horizontal scroll bar does not work with a mouse scroll wheel, too.

I think this requires more design consultation. I'd like to close this PR in favor of having a discussion with design.

jtarchie:
Discuss here concourse/concourse#946

PR #175, opened 2017-06-26
Skip builds from paused jobs when getting next pending.

jmcarp:
At the moment, `Job.GetNextPendingBuild` doesn't check whether the build it gets is part of a pending job or not. That means that pending builds of paused jobs block builds of other jobs in the same serial group, as described in https://github.com/concourse/concourse/issues/520. This patch ignores builds of pending jobs in this method.

cc @cnelson


chendrix:
Looks good @jmcarp is this ready to be merged or did you have any more work you wanted to do on it?

jmcarp:
Thanks for taking a look @chendrix ! Unit tests are passing, and I verified that pausing works with serial groups as expected with this change, so I'm happy with it unless you all have any concerns.

PR #177, opened 2017-07-02
[Bug fix] Filter resource versions by type.

jmcarp:
If the type of a resource changes, resource versions from the previous
resource type can be fetched and passed to builds. This leads to
unexpected build errors, since resource version metadata from one
resource type may not be usable to another resource type. This commit
filters resource versions by the current type of the resource.

cc @chendrix


jmcarp:
On second thought, it looks like this would only fix the versioned resources endpoint and wouldn't prevent versioned resources with types that don't match the current type of the resource from being used in builds. I can leave this open as a bug report or close and refile as an issue--WDYT?

chendrix:
Please close and refile as issue. There should be a general story somewhere about all of the problems with "changing resources" that I don't think has been made.

PR #181, opened 2017-07-12
Present auth config to admin users on listing teams.

jmcarp:
Based on chat with @chendrix around https://concourseci.slack.com/archives/C07RY25QF/p1499885468923378. The idea is for `ListTeams` to behave differently depending on the current user's access, like `ListPipelines`: it shows redacted teams if the user is not authenticated or authenticated to a non-admin team, and shows non-redacted teams if the user is authenticated to the admin team.

cc @cnelson

### DIFF
--- a/api/present/team.go
+++ b/api/present/team.go
@@ -11,3 +11,12 @@ func Team(team db.Team) atc.Team {
 		Name: team.Name(),
 	}
 }
+
+func TeamWithAdmin(team db.Team) atc.Team {
+	return atc.Team{
+		ID:        team.ID(),
+		Name:      team.Name(),
+		BasicAuth: team.BasicAuth(),
+		Auth:      team.Auth(),
+	}
+}
--- a/api/teams_test.go
+++ b/api/teams_test.go
@@ -113,6 +113,65 @@ var _ = Describe("Teams API", func() {
 					}
 				]`))
 			})
+			Context("when authenticated as non-admin team", func() {
+				BeforeEach(func() {
+					jwtValidator.IsAuthenticatedReturns(true)
+					userContextReader.GetTeamReturns("fake-team", false, true)
+				})
+				It("returns the teams with auth configuration", func() {
+					body, err := ioutil.ReadAll(response.Body)
+					Expect(err).NotTo(HaveOccurred())
+
+					Expect(body).To(MatchJSON(`[
+						{
+							"id": 5,
+							"name": "avengers"
+						},
+						{
+							"id": 9,
+							"name": "aliens"
+						},
+						{
+							"id": 22,
+							"name": "predators"
+						}
+					]`))
+				})
+			})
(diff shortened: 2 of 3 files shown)


jtarchie:
@chendrix, this seems that it might be a security related. Exposing credentials in any capacity seems bad.

We are discussing on concourse/concourse#1209 about opening pipelines for other users. Does this fit this use case?

chendrix:
@jtarchie Members of the main team can already override anyone else's auth. Is there anything strictly worse about the main team being able to see those auth credentials?

I guess if they are sharing passwords, then it's worse to be able to read the credentials vs. override them. Any others?

jmcarp:
For what it's worth, I don't actually need to see the credentials for the use case I'm looking at--I need to retrieve what are (to me) non-sensitive config values like the cf url and space guid(s). My use case is rotating credentials, but I need to know all the values in the config to update the secret parts. So it might make sense to revise this so that each provider can expose non-sensitive details to admins, or to take a different approach and allow set-team to change the value of one attribute without having to know the values of other attributes.

ebabani:
Had to investigate an issue a few weeks ago where `fly login` was failing. It turned out to be a bad auth configuration.

Being able to see the other team's auth would have sped up the investigation by a lot.

jmcarp:
Ping @chendrix and @jtarchie when you have time. No rush, just a reminder 😄.

jtarchie:
I'm still of the opinion that exposing any credentials is bad practice. Compare it to "Forgot your Password" with a website. Would you want to get your original password in plain text?

jmcarp:
@jtarchie: I understand, and I wrote up some alternatives above:

* Expose all auth config to admins (which has the security issues that you pointed out)
* Expose all non-sensitive auth config to admins (e.g., all uaa config values except the client secret)
* Allow partially overrides of auth config (e.g., override client credentials without specifying space guids)

Any of those options would work for my use case. WDYT?

chendrix:
Also if you look at the milestone, it's targeted to 3.4.0. We're still getting 3.3.x versions out the door.

cnelson:
@jtarchie @chendrix.  Could we get some feedback on an acceptable way to proceed?  We'd like to see this land in 3.4.0 as this is blocking a out ability to enable [encryption](https://concourse.ci/encryption.html) for https://github.com/18F/concourse-broker

cnelson:
@jtarchie @chendrix.  Could we get some feedback on an acceptable way to proceed?  We'd like to see this land in 3.4.0 as this is blocking a out ability to enable [encryption](https://concourse.ci/encryption.html) for https://github.com/18F/concourse-broker

topherbullock:
Closing this, as the related issue (concourse/concourse#1167) has been closed in favour of more fine grained / RBAC to achieve similar end goals. concourse/concourse#1317

PR #185, opened 2017-07-21
Utilize a proxy when configured for OAuth exchange

bkeyser:
Addresses issue described in concourse/concourse#1337 where the token exchange fails when using a proxy.

### DIFF
--- a/auth/genericoauth/provider.go
+++ b/auth/genericoauth/provider.go
@@ -177,6 +177,7 @@ func (provider Provider) Client(ctx context.Context, t *oauth2.Token) *http.Clie
 func (Provider) PreTokenClient() (*http.Client, error) {
 	return &http.Client{
 		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
 			DisableKeepAlives: true,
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@bkeyser Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=185)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=185) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bkeyser Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=185)!

jtarchie:
Thanks for the PR! I think that you've been able to narrow down the issue. I think we were assuming that `http_*` vars got passed through because that is the default beahviour with an undefined client in golang. However we are using a custom client, which needs to explicitly define these values.

Let me see if I can test this locally to accept.

PR #189, opened 2017-07-31
Added basic kubernetes secrets support

farcaller:
This adds a very basic secrets backend via kubernetes Secrets API.

Secrets are resolved in the following order:
- look up `${TeamName}/${PipelineName}-concourse-secrets`, fetch key `${varDef.Name}`
- look up `${TeamName}/${varDef.Name}`, return map of all keys


jtarchie:
@farcaller, I think we'll have to discuss the `${DefaultNamespace}/${varDef.Name}` lookup. This is because concourse deployments can be shared across multiple teams. This means teams can get access to global values. 

Also `${DefaultNamespace}/${PipelineName}-concourse-secrets` means if two teams have a name of a pipeline be the same there could be collisions.

Can you clarify when these namespaces are required? I totally acknowledge this could be my lack of knowledge of k8, so please feel free to correct me. :-P

farcaller:
Kubernetes provides isolation via [namespaces](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/). In a nutshell, k8s namespace is a completely distinct environment, with its own set of objects, including Secrets.

To get an object, you need to specify its name and namespace.

I guess that removing the DefaultNamespace and always sticking to team as namespace name would provide enough isolation?

topherbullock:
@farcaller Yeah, I _think_ setting up the secrets for a team under a namespace makes sense in terms of Kubernetes abstractions, but I'd like to see some more little feedback from others on how they deploy and use Kubernetes and Concourse ( in general for future support, and for this integration )

farcaller:
Tentatively summoning @vyshane, @frodenas, @william-tran, @jtarchie and @prydonius as having some k8s and concourse experience.

Sorry for a mass-ping, but I want to see this PR going through and I don't know any other people using concourse within k8s. I randomly sampled your names from github (and concourse helm chart).

Can you please review the suggested patch and see if it makes sense? The idea here is to allow to use k8s `Secret`s as a variable source for concourse so that concourse would run completely free from secret data in the pipelines, instead managed by the serviceaccount atc runs under.

codesuki:
Good idea, I hope this gets released soon!

We are using Concourse on GKE in its own cluster. From there we build our images and deploy them on other clusters. We only have one Concourse team which is our GitHub organization.
I think it would be good if the namespace could be configured somehow because in the concourse cluster everything runs in the same namespace anyway.

farcaller:
I'd prefer a more flexible syntax to define variables in the concourse pipelines, but I work with what I've got, and what I've got doesn't allow any nesting.

Maybe to make it slightly more flexible I could introduce a flag pointing at a ConfigMap in the same NS as ATC, and the configmap would map concourse teams to namespaces – this would allow at least some flexibility.

farcaller:
So, a month passed, and there's literally zero interest from the community. I couldn't care less if concourse would be easier to build, I'd just run it off a private fork forever (just like I do it with fly already — because https://github.com/concourse/fly/pull/150 was just closed). Sadly, it doesn't seem that obvious what constitutes a proper atc docker image.

Can we merge this as "experimental" and refine the API later on? Most of the major opensource products use a notion of fragility of API. No one will ever come to this PR to discuss the interface, I believe, not until this is merged, published in docs and people come in to try this code and comment on their experiences.

That's your project, obviously, and you set the rules. I just want to get a clearer expectation of what to expect for my further contributions (I don't like writing code that solves actual use cases that gets discarded).

jtarchie:
@farcaller, it has not been neglected. We don't user or understand kubernetes features. This is why we were waiting for community feedback.

We appreciate the pull request. For future endeavors, it is customary to create an issue with a feature request, so discussion can be made.

We don't want to merge in experimental credential features because it has the potential of security issue -- i.e. leaking credentials. This has nothing to do with the API, just the current implementation, which we need validation from the community to support.

We've reached out to people on slack and individually for help on k8. We'd like to understand its usage and needs, but need help via someone who is an *expert/champion*, not us learning from scratch. This has to do with our limited time to work on our current prioritized feature set. Would you be able to help organize the community to help us with that?

vito:
@timrchavez @william-tran Do either of you have experience with the k8s secrets API? We need some assistance on this one. (Thanks!)

william-tran:
Ergin and I will review this today.

william-tran:
I had a look at this with @badie, from a code perspective the error return could be better: https://github.com/concourse/atc/pull/189/files#diff-3bd6ba02f2cd8824de859ce43afb2ea1R53 and the comment for `List` https://github.com/concourse/atc/pull/189/files#diff-3bd6ba02f2cd8824de859ce43afb2ea1R66 could just be "unsupported operation". But those are minor code style issues and the concourse team would be better judges of that. 

From a k8s perspective, the client can be set up for [in-cluster API access](https://github.com/kubernetes/client-go/tree/master/examples/in-cluster-client-configuration): https://github.com/concourse/atc/pull/189/files#diff-714fbfe2f42ad21d1a0a2b90d4306512R31

as well as [out of cluster](https://github.com/kubernetes/client-go/tree/master/examples/out-of-cluster-client-configuration), where concourse will use a client identity from a config file
https://github.com/concourse/atc/pull/189/files#diff-714fbfe2f42ad21d1a0a2b90d4306512R27

Access to secrets by either of these identities can be controlled by [k8s RBAC](https://kubernetes.io/docs/admin/authorization/rbac/). Rules can be configured to allow the concourse identity GET access to all secrets in each namespace/team. Given a service account `atc` assigned to pod `concourse-web` in the `concourse` namespace, the following would grant access to the secrets for team `main`:

```
# https://kubernetes.io/docs/admin/authorization/rbac/#role-and-clusterrole
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: secrets-get
rules:
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["get"]

---

# https://kubernetes.io/docs/admin/authorization/rbac/#rolebinding-and-clusterrolebinding
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: concourse-get-main-secrets
  namespace: main
subjects:
- kind: ServiceAccount
  name: concourse:atc
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: secrets-get
  apiGroup: rbac.authorization.k8s.io

```

The lookup rules could be improved though. What if the cluster already has a namespace called `main`?  To map team names to namespaces, it would be more obvious and reduce the chance of name collision if we did namespace = concourse-team-[team_name]. The names of secrets must be valid hostnames, and namespaces must be valid lables (no dots), so there needs to be some sanitization on team and pipeline names to be used as k8s identifiers. Secrets themselves are key-value maps like Vault. So I'm proposing the lookup rules to be:

In team `accounting_dev`, pipeline `boring_app`, reference `((api-key))`:

1. The value in namespace: `concourse-team-accounting-dev` secret: `boring_app.api-key`, key: `value`
2. The value in namespace: `concourse-team-accounting-dev` secret: `api-key`, key: `value`

In team `accounting_dev`, pipeline `boring_app`, reference `((common-secrets.api-key))`:

1. The value in namespace: `concourse-team-accounting-dev` secret: `boring_app.common-secrets`, key: `api-key`
2. The value in namespace: `concourse-team-accounting-dev` secret: `common-secrets`, key: `api-key`

ebabani:
Thanks for this PR @farcaller  . This would definitely be something we'd use on our k8s cluster.

I agree with the secret lookup @william-tran is proposing above. It's consistent with how Vault and Credhub secrets are handled. 

For namespace collisions if we use the team name as the kubernetes namespace, how about adding an optional `--kubernetes-secret-namespace-prefix` setting? The namespace lookup would be `[prefix][team-name]`, with the default prefix being empty so you use just the team name.

farcaller:
I agree with the proposed lookup rules and can refactor the code as suggested.

vito:
Careful with sanitizing to determine the namespace. If I know there's a team `foo_bar`, I may be able to create another Concourse team `foo-bar` and get all their secrets.

william-tran:
@vito good catch.

farcaller:
So, I'm hacking into the proposed solution and I'm having troubles with the

> In team accounting-dev, pipeline boring-app, reference ((api-key)):
> In team accounting-dev, pipeline boring-app, reference ((common-secrets.api-key)):

as there's no way to figure if the api requested `((api-key))` or `((common-secrets.api-key))` as far as I see it, more so, it seems a bit too complex.

What if we do this instead?

In team `accounting-dev`, pipeline `boring-app`, reference `((api-key))`:

Look up secret `prefix+accounting-dev/boring-app`, look up `api-key` in the secret KV, return
if it fails:
Look up secret `prefix+accounting-dev/boring-app.api-key`, return all the keys.

ebabani:
When would we want to return all keys? I don't think I've seen that with Vault ever. 

With your above proposal how would multiple pipelines sharing the same secret work?
eg. I want secret `((github_key))` for pipelines `p1` and `p2`. Where would I need to create the kubernetes secret for both pipelines to be able to use it?

farcaller:
Vault does it here: https://github.com/concourse/atc/blob/master/creds/vault/vault.go#L46 I think it happens when you don't have the "value" key?

> Vault credentials are actually key-value, so for ((foo)) Concourse will default to the field name value. You can specify the field to grab via . syntax, e.g. ((foo.bar)).

It's very obscure because Concourse never passes the `bar` part to the backend, so the backend needs to guess what the user was asking for.

william-tran:
@farcaller How about this for the implementation, which should provide the same result as I described (though in JS):

```js
function getCredentials(varDef) {
  var namespace = prefix + teamName;
  var secret = k.findSecret(namespace, pipelineName + '.' + varDef.Name) || k.findSecret(namespace, varDef.Name);

  if (!secret) {
    throw new Error("couldn't find a matching secret");
  }
  
  // if there is a key in secret called "value", return its value
  if (secret['value']) {
    return secret['value'];
  }

  // otherwise return the entire map, assuming the user asked for ((foo.bar))
  return secret;
}

```

vito:
thanks all! we've merged #225 now.

PR #194, opened 2017-08-29
CredHub integration with ATC

leungandrew:
This is what an implementation of using CredHub credential management in Concourse looks like.  It now integrates using the new CredHub Go library at github.com/cloudfoundry-incubator/credhub-cli.

This has only been tested manually by running a local instance of CredHub and Concourse, so tests should be written in TopGun afterwards.

Dependencies have been vendored using govendor.


pivotal-issuemaster:
@leungandrew Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=194)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=194) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@leungandrew Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=194)!

creds/credhub/manager.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+			manager.caCerts = append(manager.caCerts, string(contents))
+		}
+	}
+
+	return nil
+}
+
+func (manager CredHubManager) NewVariablesFactory(logger lager.Logger) (creds.VariablesFactory, error) {
+	var skipTls credhub.Option
```
leungandrew: I'm not thrilled with the way the credhub api has been implemented w.r.t. the SkipTLSValidation option.  I'd consider making a pull request to change the API to make it easier to use (perhaps it should accept a boolean as a parameter).  If there's a better way to implement this though I'm open to suggestions.

vito:
fyi: I requested that we vendor the dependencies for this as I've liked the pattern previously as it makes extension points like this truly self-contained, and easy to version.

However now looking at the tree I'm seeing a lot more than I expected/wanted - I wonder if the vendor tooling is including the entire repo and its dependencies rather than just the packages we use (and the packages they use and so on)?

Maybe we should just not vendor; I'm afraid that we'll get into a state where we depend on different versions of the same package (e.g. `golang/x/crypto`), which may be painful.

leungandrew:
Yeah, of the credhub code itself, I only pulled in the package of the API, but the API itself has their own dependencies and govendor pulled those in as well.  Let me know if you'd like me to remove the vendoring.

jtarchie:
Merged! We are going to keep this open until we finish.

* [ ] add bosh properties
* [ ] document credhub usage

PR #195, opened 2017-08-31
Add web app manifest handler.

headcr4sh:
### Background
For my use case, I would like to have a web app manifest, because it allows me to "pin" the website of my Concourse CI instance onto the homescreen of my Android tablet and enjoy a "borderless" view of the pipelines overview.

### Further steps
Providing a web app manifest is the first useful step towards providing a real app-like experience on devices where such a behavior is applicable/desirable. Stuff like "responsive design" and the use of "service workers" to provide a truly compelling "progressive web app" are things that might be added in the future...

### What are these Web App Manifests anyhow?
For further details about web app manifests in general see:
* https://www.w3.org/TR/appmanifest/
* https://developer.mozilla.org/en-US/docs/Web/Manifest
* https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/


web/manifest/handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+type handler struct {
+	m manifest
+}
+
+// NewHandler creates a new handler that is able to serve web app manifest.
+func NewHandler() Handler {
+	return &handler{
+		m: manifest{
+			ShortName:       "Concourse",
```
headcr4sh: All this stuff is hard-coded which is hopefully o.k. for now.

The ability to change certain parameters may be added in the future, though-- that's why I decided to go with a "possibly dynamic approach" of creating the web app manifest instead of dropping a static JSON file into the "/public/" folder.

pivotal-gabriel-dumitrescu:
Thanks @headcr4sh for the PR.  We tried it on a Nexus tablet: 

steps:
- create shortcut on the home screen, by using the Chrome menu option on the top right side
- tap on the icon shortcut on the home screen

expected: 
- page to open and get a real app-like experience on device

actual: 
- page was opened in the Chrome browser, showing the normal browser experience on mobile device  

Can you let us know if there are any other steps maybe we missed or we should take to reproduce the correct state?

headcr4sh:
Hi there.
I must have missed some things whilst preparing my PR. Sorry...
The manifest information is correctly served, but for some reason the updated contents which I put into index.html no longer works (I am pretty sure, it _did_ work, though,...).
I am going to look into it.

Btw.:
I have a Nexus 9 tablet as well which I want to use as dashboard on my desk. :-)

headcr4sh:
I have added the missing necessary files that are needed to allow the "pin to home screen" functionality on Android devices. It now works as expected on my very own Nexus 9 device.

--- I ammended my changes onto the previous commit, hope that's the expected workflow ---

Some more notes
* I had to re-create the binary blob "web/bindata.go"
* An icon of exactly 144x144 px seems to be necessary, thus I added one
* I also added an SVG icon to make sure that the the icon displays nicely on hi-res displays

Out-of-scope:
* Chrome for Desktop (on my Linux box at least) seems to support a feature where users can add web apps to the user's desktop. I ran into an error though, because on desktop systems a service worker seems to be necessary.

pivotal-gabriel-dumitrescu:
thanks!

PR #198, opened 2017-09-28
Error if multiple emitters are configured.

jmcarp:
If multiple emitters are configured, only one is used; the others are
silently ignored. This patch causes the atc to fail if multiple emitters
are configured so that users don't expect that multiple emitters are
supported.

[Closes https://github.com/concourse/concourse/issues/1557]

h/t @jtarchie


vito:
thanks!

PR #199, opened 2017-09-28
Add option to use LetsEncrypt to obtain TLS certificates

takeyourhatoff:
Adds the option `--autocert` which can be used instead of `--tls-cert` and `--tls-key`.

If `--autocert` is given, ATC will generate a key and certificate, then, using the ACME protocol, get the certificate signed and start using it to serve HTTPS. The ACME provider defaults to LetsEncrypt but can be changed with the `--acme-url` option.

When `--autocert` is used, ATC will handle certificate expiration by renewing the certificate it has previously obtained when that certificate's expiry date is less than 30 days time. This happens without impacting the availability of the server.

ATC will cache certificates it has obtained using ACME in the database table `cert_cache`. This done because LetsEncrypt implements severe rate limiting on obtaining duplicate certificates. Only 5 certificate with the same set of requested hostnames will be issued in any given 7 day period.

Fixes [#918](https://github.com/concourse/concourse/issues/918)

### DIFF
--- a/atccmd/cert_cache.go
+++ b/atccmd/cert_cache.go
@@ -1,0 +1,61 @@
+package atccmd
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/concourse/atc/db"
+	"golang.org/x/crypto/acme/autocert"
+)
+
+type dbCache struct {
+	get, put, delete *sql.Stmt
+	es               db.EncryptionStrategy
+}
+
+func newDbCache(conn db.Conn) (autocert.Cache, error) {
+	c := new(dbCache)
+	c.es = conn.EncryptionStrategy()
+	var err error
+	c.get, err = conn.Prepare("SELECT data, nonce FROM cert_cache WHERE key = $1")
+	if err != nil {
+		return nil, err
+	}
+	c.put, err = conn.Prepare("INSERT INTO cert_cache (key, data, nonce) VALUES ($1, $2, $3)")
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@takeyourhatoff Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=199)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=199) to manually synchronize the status of this Pull Request.

jmcarp:
Just my $0.02, but I agree with points from @chendrix and @topherbullock in https://github.com/concourse/concourse/issues/918. I think this functionality belongs in a generic proxy like traefik or caddy. I haven't used this, but I see there's a bosh release that supports auto-tls with caddy at https://github.com/govau/instant-https-boshrelease.

topherbullock:
Yeah, I tend to be on the same page as @jmcarp here, concourse/concourse#918 was closed because we didn't want to add that additional thing for people to configure, and want to keep the generation of certs away from Concourse. There should be a way to do this without duplicating the effort on Concourse ( / any thing that uses a web server ) to integrate with LetsEncrypt.

It's doubly concerning that not only does this go out and get the certs from somewhere, but it has to store them in the database. Yes they're encrypted and it makes sense why, but it further points to the amount of effort we'll need to put in to maintain this.

PR #184, opened 2017-07-21
Finalised token implementation for the login

Chumper:
@jtarchie 
@vito 
@chendrix 

Please review & merge.
This is linked to https://github.com/concourse/fly/pull/185 and needs to be merged first.

This PR enables the ATC to return the following structure on `api/v1/teams/:team_name/auth/methods`:

```    
{
        "auth_url": "http://127.0.0.1:8080/auth/github?team_name=openCloak",
        "display_name": "GitHub",
        "token_url": "http://127.0.0.1:8080/auth/github/token?team_name=openCloak",
        "type": "oauth"
}
```

Note the added `token_url`

PR #202, opened 2017-10-03
First draft of LDAP authentication

dthorsen:
I have been working on implementing LDAP authentication in Concourse and have a basic working implementation. I need to update tests and add new tests.  I also need to finish adding the command flag validation.  But I wanted to get what I have done so far out here to solicit some feedback on how I implemented this. Would someone have some time to look this over and let me know if there are any improvements or changes in approach desired for this feature?

Thanks in advance!

### DIFF
--- a/api/teamserver/set.go
+++ b/api/teamserver/set.go
@@ -122,5 +122,10 @@ func (s *Server) updateCredentials(atcTeam atc.Team, team db.Team) error {
 		return err
 	}
 
+	err = team.UpdateLdapBasicAuth(atcTeam.LdapBasicAuth)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/atccmd/command.go
+++ b/atccmd/command.go
@@ -862,6 +862,29 @@ func (cmd *ATCCommand) configureAuthForDefaultTeam(teamFactory db.TeamFactory) e
 		return err
 	}
 
+	var ldapBasicAuth *atc.LdapBasicAuth
+	if cmd.Authentication.LdapBasicAuth.IsConfigured() {
+		ldapBasicAuth = &atc.LdapBasicAuth{
+			Server:                cmd.Authentication.LdapBasicAuth.Server,
+			Port:                  cmd.Authentication.LdapBasicAuth.Port,
+			TLSEnabled:            cmd.Authentication.LdapBasicAuth.TLSEnabled,
+			TLSInsecureSkipVerify: cmd.Authentication.LdapBasicAuth.TLSInsecureSkipVerify,
+			TLSCA:         cmd.Authentication.LdapBasicAuth.TLSCA,
+			BindUsername:  cmd.Authentication.LdapBasicAuth.BindUsername,
+			BindPassword:  cmd.Authentication.LdapBasicAuth.BindPassword,
+			UserBaseDN:    cmd.Authentication.LdapBasicAuth.UserBaseDN,
+			GroupBaseDN:   cmd.Authentication.LdapBasicAuth.GroupBaseDN,
+			GroupDN:       cmd.Authentication.LdapBasicAuth.GroupDN,
+			UserAttribute: cmd.Authentication.LdapBasicAuth.UserAttribute,
+			GroupFilter:   cmd.Authentication.LdapBasicAuth.GroupFilter,
+		}
+	}
+
+	err = team.UpdateLdapBasicAuth(ldapBasicAuth)
+	if err != nil {
+		return err
+	}
+
 	teamAuth := make(map[string]*json.RawMessage)
 	for name, config := range cmd.ProviderAuth {
 		if config.IsConfigured() {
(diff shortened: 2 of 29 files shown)


pivotal-issuemaster:
@dthorsen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=202)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=202) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dthorsen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=202)!

dthorsen:
Just to link this to the related issue/feature request. This would add support for LDAP as requested in this issue: #338

vito:
thanks for taking this on! this may take a bit for us to process though, which isn't at all your fault. it's a factor of:

* we don't know much about LDAP
* we generally treat auth PRs with more scrutiny
* i'm concerned that his had to touch a bunch of different parts of the codebase - we were able to at least consolidate oAuth providers into a generic interface, guaranteeing that all necessary parts of the code can be PRed at once, but this looks like a new beast

it's always bothered me that we let basic auth be a special case - i wonder if this, plus learnings from #215, could illustrate what some new aspects of a new generic interface could look like?

dthorsen:
Thanks @vito for the feedback! This is indeed a new beast. The thing about LDAP is that from a client perspective, it is still HTTP basic auth, so must flow through the same workflow from a UX perspective. It simply needs replace the credential verification implementation with one that checks against LDAP instead of against the database. An additional challenge with the LDAP implementation is that it should ideally be flexible enough to work with any RFC compliant LDAP implementation, not just Active Directory, though I suspect in the majority of cases this would be used with Active Directory. The approach in this PR was to sort of extend BasicAuth to also support using an LDAP backend (when configured).

A generic authentication interface would be very nice. I have given a little thought as to what that could look like, and it would require some pretty major surgery on the existing authentication code. To be thorough, it would probably require nesting multiple interfaces due to the varying UX required for different authentication workflows.

vito:
@dthorsen Thanks for the context.

This may sound a bit heavy but I think it may be wise to look into having this fit the general "provider" mold after all. I'm concerned that this follows a precedent that we don't want to continue (having another auth flow that touches many risky parts of the codebase). This may be a good opportunity to further generalize the [`provider.Provider`](https://github.com/concourse/atc/blob/master/auth/provider/provider.go#L19-L30) interface to support more than just oAuth2.

Note that everything else in the `provider` package is fairly general, and would easily apply to basic auth, LDAP auth, and just about any type of auth.

Here's one approach: rather than having special handlers registered in `command.go` for [oAuth](https://github.com/concourse/atc/blob/master/atccmd/command.go#L387-L397) and hard-coding basic auth validation into [the Get Token endpoint](https://github.com/concourse/atc/blob/894fc604a6586bd225276c0708154104738c6f09/auth/get_token_validator.go#L32-L34), auth providers could somehow set up their own handlers? For example, the basic auth provider could register an endpoint that verifies the basic auth, and the oAuth providers could all register an oAuth handler?

To be honest I think this'll take some digging around in the code and just chasing types around for a bit, but I think it's worth exploring. With this auth flow blazing the trail, we could then switch the basic auth provider over ourselves (since that would involve a costly migration that I doubt you'd want to risk messing up). 😄

dthorsen:
@vito That approach makes sense. Will need to flesh out a bunch of details. I will take a stab at generalizing the interface `provider.Provider` interface a bit. For first pass I will probably just have the LDAP provider stub some of the oAuth implementation to satisfy the existing interface (since I don't want to modify all the oAuth providers in this PR).

Somehow we will also need to deal with the change in UX flow as well. Since LDAP will require the web login form to collect username/password, but oAuth does not. I am not sure what the current concourse UX is at CLI for oAuth login. I will check that too to see what changes might be needed for the CLI UX.

I am also trying to figure out how this could be done in a transitional way so that all the oAuth providers continue to function until they are transitioned to a more generalized interface.

vito:
@dthorsen One way to tackle the login form challenge may be to have providers able to register their own handlers under `/auth/XXX`. The LDAP and basic auth providers could then register one that does the basic auth prompt. The oAuth providers would no longer be specialized [here](https://github.com/concourse/atc/blob/8ae891ee0356131b221cfc9c8296f38a2928221c/auth/routes/oauth_routes.go#L13-L14) and would register handlers under their own name.

One thing that's kinda nice about this is that the `atc.AuthMethod` that they provide would then refer directly to the endpoint that the provider itself registered, keeping it more self-contained.

This kind of flow may also have been useful for the previous attempt at GitHub token-based auth  (#184).

Does that make sense?

I'm not sure about how `fly` would know to ask for username/password though, and for the oAuth case how it would know to do the fancy click-URL-to-log-in flow. Handlers may differ in what values they expect, even.

I do want to point out that we're likely solving the exact problems that [UAA](https://github.com/cloudfoundry/uaa) solves. That's not to say we should just switch to it (probably no one wants to deploy it for Concourse alone), but there may be patterns to steal. For example, I think their auth methods endpoint returns a list of values to prompt for. But we could also just have `fly` magically know. ;) I can't imagine how we could generalize the fancy callback flow.

vito:
Closing this in light of concourse/concourse#1888. We'll keep in touch if we back out of that plan.

Thanks again!

PR #205, opened 2017-10-10
Resolution for 1693. Base64 encode the ATC-Authorization cookie value

brandon-bethke-neudesic:
This pull request will base64 encode/decode the ATC-Authorization cookie value. It provides a resolution for Issue #1693 concourse/concourse#1693

### DIFF
--- a/api/authserver/get_auth_token.go
+++ b/api/authserver/get_auth_token.go
@@ -1,6 +1,7 @@
 package authserver
 
 import (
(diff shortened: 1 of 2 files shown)


vito:
Hmm it sounds like they've fixed the root of the bug stream. I don't really want to merge this since it sounds like it should no longer be necessary, and it's a bit confusing as to why it'd need to be there otherwise. Thanks though!

PR #207, opened 2017-10-17
We should provide an auth flag when running atc

mumoshu:
The `go run` command fails like this without a flag:

```
$ go run cmd/atc/*.go
1 error occurred:

* must configure basic auth, OAuth, UAAAuth, or provide no-auth flag
exit status 1
```


vito:
thanks!

PR #208, opened 2017-10-17
Actually use credential management variables

aeijdenberg:
It's not clear to me how anyone has managed to get CredHub integration to work without this fix?


aeijdenberg:

PR #209, opened 2017-10-17
Fix panic that occurred when "insecure-skip-verify" was not set.

aeijdenberg:
Previously an "empty" function was passed to credhub.New() which promptly panicked.

Change error handling to fail fast if an error is returned.


aeijdenberg:

PR #210, opened 2017-10-17
Change IsConfigured to OR instead of AND for CredHub

aeijdenberg:
(if the user specifies any option, this indicates intent)

Add and simplify validation for CredHub


aeijdenberg:

PR #211, opened 2017-10-17
Return an error if an unrecognized positional argument is returned.

aeijdenberg:
For example, `--credhub-insecure-skip-verify false`

(boolean flags don't take values, so false is ignored)

 (the bosh release currently generates: `--credhub-insecure-skip-verify true`, I haven't checked to see what happens if `false` is specified, that's my next PR)


aeijdenberg:

PR #213, opened 2017-10-17
Fixup a number of issues with the credhub integration

aeijdenberg:
In experimenting with the CredHub integration I came across a number of issues while debugging.

Once this PR passes CI, I'll abandon the separate PRs I made for these earlier today.

See also related https://github.com/concourse/concourse/pull/1725.

### DIFF
--- a/acceptance/atc_command_test.go
+++ b/acceptance/atc_command_test.go
@@ -14,7 +14,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"time"
 
 	"github.com/tedsuo/ifrit"
@@ -335,7 +334,7 @@ func (a *ATCCommand) getATCCommand() *exec.Cmd {
 	}
 
 	if a.telemetryOptIn {
-		params = append(params, "--telemetry-opt-in", strconv.FormatBool(a.telemetryOptIn))
+		params = append(params, "--telemetry-opt-in")
 	}
 
 	return exec.Command(a.atcBin, params...)
--- a/atccmd/command.go
+++ b/atccmd/command.go
@@ -208,7 +208,10 @@ func (cmd *ATCCommand) Execute(args []string) error {
 	return <-ifrit.Invoke(sigmon.New(runner)).Wait()
 }
 
-func (cmd *ATCCommand) Runner(args []string) (ifrit.Runner, error) {
+func (cmd *ATCCommand) Runner(positionalArguments []string) (ifrit.Runner, error) {
+	if len(positionalArguments) != 0 {
+		return nil, fmt.Errorf("unexpected positional arguments: %v", positionalArguments)
+	}
 	err := cmd.validate()
 	if err != nil {
 		return nil, err
--- a/atccmd/command.go
+++ b/atccmd/command.go
@@ -192,7 +192,7 @@ func (cmd *ATCCommand) WireDynamicFlags(commandFlags *flags.Command) {
 
 	managerConfigs := make(creds.Managers)
 	for name, p := range creds.ManagerFactories() {
-		managerConfigs[name] = p.AddConfig(authGroup)
+		managerConfigs[name] = p.AddConfig(credsGroup)
 	}
 	cmd.CredentialManagers = managerConfigs
 
--- a/creds/credhub/credhub.go
+++ b/creds/credhub/credhub.go
@@ -72,7 +72,7 @@ func (c CredHubAtc) findCred(path string) (credentials.Credential, bool, error) 
 		return cred, false, nil
 	}
 
-	cred, err = c.CredHub.Get(path)
+	cred, err = c.CredHub.GetLatestVersion(path)
 	if _, ok := err.(*credhub.Error); ok {
 		return cred, false, nil
 	}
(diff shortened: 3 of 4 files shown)


aeijdenberg:
@vito - thanks for merging the PR in the Concourse bosh release yesterday. Not sure if you saw this one or not, but wanted to draw it to your attention as you'll need to take at least some of the commits in this PR too before cutting a new bosh release, else anyone who isn't explicitly setting `--insecure-skip-verify` will see a nasty `panic` when `atc` starts up.

jtarchie:
@aeijdenberg, can you be a little more explicit about your changes here. Assume they are not obvious.

aeijdenberg:
Hi @jtarchie,

Here are some details:

1. Changes `atc` to fail validation of arguments at startup if an unexpected positional argument is specified. This is to prevent a re-occurrence of a bug like <https://github.com/concourse/concourse/pull/1725>, whereby if `atc` is invoked with:

```bash
atc ... --credhub-insecure-skip-verify false
```

then this would actually have the opposite behavior than desired. ie since it's a boolean parameter, no value is expected, just specifying it sets it to `true`, and any value passed, prior to the change, was silently ignored by `atc`.

2. The latest version of `github.com/cloudfoundry-incubator/credhub-cli/credhub` which seems to be installed on your CI server, has some slightly changed method names, so this updates to compile. i.e `Get()` changes to `GetLatestVersion()`, and `SkipTLSValidation()` now takes a parameter.

3. `IsConfigured()` was previously incorrectly requiring that all parameters were set (and it was missing some). Changes this to check if *any* parameters are set, so that the `Validate()` method is invoked any time the user specifies intent to use CredHub.

4. Changes `Validate()` to actually validate all required arguments, instead of a subset.

5. Fix `NewVariablesFactory()` so that it stops passing a `nil` `credhub.Option` which would cause a `panic()` on start up any time that `--credhub-insecure-skip-verify` was not set.

6. Adds the credential manager set of options to the `credsGroup` instead of the `authGroup`. I'm less sure of this one - it appeared to me that this is more cosmetic fix.

I suspect there are probably more issues, but these changes at least allow `atc` to startup successfully when `--credhub-insecure-skip-verify` is not set, which was previously (a) not possible due to the `panic()` and (b) difficult to invoke due to the other ERB generation bug.

My suspicions is that we'll need some extra parameters to include a CA cert for UAA, which may have a different CA than CredHub (current that's the default behavior for a `bosh` install, see <https://github.com/cloudfoundry/bosh-deployment/pull/157>), but I hadn't gotten to that point of testing yet.

Hope that helps.

aeijdenberg:
@vito @jtarchie 

I note that v3.6.0 Concourse is now released, without this patch. I've tested it locally and verified that a previously valid configuration of:

```yaml
- type: replace
  path: /instance_groups/name=web/jobs/name=atc/properties?/credhub
  value:
    client_id: foo
    client_secret: bar
    url: https://127.0.0.1
    tls:
      ca_cert: |
        a cert
```

will cause a segfault on startup:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xaade3a]

goroutine 1 [running]:
github.com/cloudfoundry-incubator/credhub-cli/credhub.New(0x7ffd4f0dce19, 0x11, 0xc42004b340, 0x2, 0x2, 0xc42004b320, 0x412b08, 0x60)
	/var/vcap/packages/atc/src/github.com/cloudfoundry-incubator/credhub-cli/credhub/new.go:29 +0x10a
github.com/concourse/atc/creds/credhub.CredHubManager.NewVariablesFactory(0x7ffd4f0dce19, 0x11, 0x7ffd4f0dceb1, 0xa, 0xc4201f0790, 0x1, 0x1, 0x0, 0x7ffd4f0dce7b, 0x3, ...)
	/var/vcap/packages/atc/src/github.com/concourse/atc/creds/credhub/manager.go:61 +0xf2
github.com/concourse/atc/creds/credhub.(*CredHubManager).NewVariablesFactory(0xc4200f8a00, 0x11363a0, 0xc420052c60, 0xc42000e298, 0x1, 0x1, 0x11363a0)
	<autogenerated>:1 +0x92
github.com/concourse/atc/atccmd.(*ATCCommand).Runner(0xc4201c8000, 0xc420173c00, 0x0, 0x6e, 0x6f, 0xc420173c00, 0x0, 0x6e)
	/var/vcap/packages/atc/src/github.com/concourse/atc/atccmd/command.go:249 +0x5ae
github.com/concourse/atc/atccmd.(*ATCCommand).Execute(0xc4201c8000, 0xc420173c00, 0x0, 0x6e, 0x0, 0x0)
	/var/vcap/packages/atc/src/github.com/concourse/atc/atccmd/command.go:203 +0x4d
main.main()
	/var/vcap/packages/atc/src/github.com/concourse/atc/cmd/atc/main.go:24 +0xe4
```

Operators can workaround by explicitly setting `insecure_skip_verify: true`, ie:

```yaml
- type: replace
  path: /instance_groups/name=web/jobs/name=atc/properties?/credhub
  value:
    client_id: foo
    client_secret: bar
    url: https://127.0.0.1
    tls:
      insecure_skip_verify: true
      ca_cert: |
        a cert
```

however without taking some of the commits in this PR, it is not possible to enable TLS validation (which was always documented as default, but only became so, and possible to enable via the bosh release in v3.6.0).

vito:
thanks a bunch!

PR #214, opened 2017-10-24
Add support for renaming teams on the API

christophermancini:
Added an ATC API endpoint for enabling teams via the Fly CLI.

See https://github.com/concourse/concourse/issues/694

### DIFF
--- a/api/teams_test.go
+++ b/api/teams_test.go
@@ -499,4 +499,115 @@ var _ = Describe("Teams API", func() {
 			})
 		})
 	})
+
+	Describe("PUT /api/v1/teams/:team_name/rename", func() {
+		var response *http.Response
+		var teamName string
+
+		JustBeforeEach(func() {
+			request, err := http.NewRequest(
+				"PUT",
+				server.URL+"/api/v1/teams/"+teamName+"/rename",
+				bytes.NewBufferString(`{"name":"some-new-name"}`),
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			response, err = client.Do(request)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		BeforeEach(func() {
+			fakeTeam.IDReturns(2)
+		})
+
+		Context("when authenticated", func() {
+			BeforeEach(func() {
+				jwtValidator.IsAuthenticatedReturns(true)
+			})
+			Context("when requester belongs to an admin team", func() {
+				BeforeEach(func() {
+					teamName = "a-team"
+					fakeTeam.NameReturns(teamName)
+					userContextReader.GetTeamReturns(atc.DefaultTeamName, true, true)
+					dbTeamFactory.FindTeamReturns(fakeTeam, true, nil)
+				})
+
+				It("constructs teamDB with provided team name", func() {
+					Expect(dbTeamFactory.FindTeamCallCount()).To(Equal(1))
+					Expect(dbTeamFactory.FindTeamArgsForCall(0)).To(Equal("a-team"))
+				})
+
+				It("renames the team to the name provided", func() {
+					Expect(fakeTeam.RenameCallCount()).To(Equal(1))
+					Expect(fakeTeam.RenameArgsForCall(0)).To(Equal("some-new-name"))
+				})
+
+				It("returns 204 no content", func() {
+					Expect(response.StatusCode).To(Equal(http.StatusNoContent))
+				})
+			})
+
+			Context("when requester belongs to the team", func() {
+				BeforeEach(func() {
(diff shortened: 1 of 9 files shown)


pivotal-issuemaster:
@christophermancini Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=214)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=214) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@christophermancini Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=214)!

api/teamserver/rename.go
```diff
@@ -0,0 +1,74 @@
...
+		return
+	}
+
+	if !found {
+		logger.Info("team-not-found")
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	fmt.Println(team.Name(), atc.DefaultTeamName)
```
vito: rogue println

api/teamserver/rename.go
```diff
@@ -0,0 +1,74 @@
...
+	if team.Name() == atc.DefaultTeamName {
```
vito: is this necessary? renaming the 'main' team should be fine - all logic is based on it being 'admin: true', not by name

christophermancini: At first my thought was we shouldn't rename it but you are right, I am not sure there is anything really preventing us from allowing it. I will remove it.

api/teamserver/rename.go
christophermancini: Good catch. Should add something to the pipeline that looks for stdout lines like that.

api/teams_test.go
```diff
@@ -499,4 +499,115 @@ var _ = Describe("Teams API", func() {
...
+					dbTeamFactory.FindTeamReturns(fakeTeam, true, nil)
+				})
+
+				It("returns 403 Forbidden", func() {
+					Expect(response.StatusCode).To(Equal(http.StatusForbidden))
+					Expect(fakeTeam.RenameCallCount()).To(Equal(0))
+				})
+			})
+
+			Context("when trying to rename the admin team", func() {
```
vito: should this be deleted now? this is no longer a special-case, right?

vito:
Build failed: https://ci.concourse.ci/builds/406885#L59f9e6ce:114:119

db/team_test.go
```diff
@@ -60,6 +60,16 @@ var _ = Describe("Team", func() {
 		})
 	})
 
+	Describe("Rename", func() {
+		JustBeforeEach(func() {
+			Expect(team.Rename("oopsies")).To(Succeed())
+		})
+
+		It("renames the team", func() {
+			Expect(team.Name()).To(Equal("oopsies"))
```
vito: This would pass even if it didn't persist the change; maybe it should test against a newly fetched team instead? Then you wouldn't have to update the `t.name` field; I'd like those fields to always come from the database and not be changed by anything but a `.Reload` method (which currently exists on other objects, but not Team).

vito:
Let's just update the test

vito:
@christophermancini Any update on this? I think we're just waiting on the test change at this point. 😄

api/teamserver/rename.go
```diff
@@ -0,0 +1,66 @@
...
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/concourse/atc"
+	"github.com/concourse/atc/auth"
+)
+
+// RenameTeam allows an authed user with authority or admin to rename a team
```
marco-m: s/authed/authenticated/      ?

christophermancini:
@vito Let me know if my changes to the test look good. If so, I can squash and update the PR.

db/team.go
```diff
@@ -121,6 +122,22 @@ func (t *team) Delete() error {
...
+	_, err := psql.Update("teams").
+		Set("name", name).
+		Where(sq.Eq{
+			"id": t.id,
+		}).
+		RunWith(t.conn).
+		Exec()
+
+	if err == nil {
+		t.name = name
```
vito: Oh, I think this can be removed; the tests shouldn't need it anymore. I'd rather leave those fields to only ever be set on first fetch, and keep the objects stateless.

christophermancini: Understood, I will remove it.

vito: (ignore this comment, just testing out the github api)

PR #215, opened 2017-10-25
Bitbucket auth

SHyx0rmZ:
This is our first attempt at adding Bitbucket auth. It's not complete yet, as we only support users at this point, but I wanted to get some feedback on the rather hacky code that makes the provider interface work with Bitbucket Server's OAuth 1 integration.


auth/bitbucket/cloud/auth_config.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/hashicorp/go-multierror"
+	"github.com/tedsuo/rata"
+)
+
+type AuthConfig struct {
+	ClientID     string `json:"client_id" long:"client-id" description:"Application client ID for enabling Bitbucket OAuth"`
+	ClientSecret string `json:"client_secret" long:"client-secret" description:"Application client secret for enabling Bitbucket OAuth"`
+
+	Users        []string                     `json:"users,omitempty" long:"user" description:"Bitbucket users that are allowed to log in"`
+	Repositories []bitbucket.RepositoryConfig `json:"repositories,omitempty" long:"repository" description:"Bitbucket repositories whose members are allowed to log in"`
```
vito: Just to confirm, this currently does nothing?

auth/oauth_callback_handler.go
```diff
@@ -73,7 +76,10 @@ func (handler *OAuthCallbackHandler) ServeHTTP(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	stateJSON, err := base64.RawURLEncoding.DecodeString(r.FormValue("state"))
+	// Read the state from the cookie instead of the param, as the param
+	// will be empty if this is an OAuth 1 request. For OAuth 2, we already
+	// made sure that the cookie and the param contain the same state.
+	stateJSON, err := base64.RawURLEncoding.DecodeString(cookieState.Value)
```
vito: Thanks for the comment here - makes sense.

auth/bitbucket/server/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	p.secrets[requestToken] = requestSecret
+	return authorizationURL.String()
+}
+
+func (p *Provider) Client(ctx context.Context, t *oauth2.Token) *http.Client {
+	return p.Config.Client(ctx, &oauth1.Token{Token: t.AccessToken, TokenSecret: t.RefreshToken})
+}
+
+func (p *Provider) Exchange(ctx context.Context, code string) (*oauth2.Token, error) {
+	r := ctx.Value("request").(*http.Request)
```
vito: Sneaking this through the context seems...sneaky. But I think I see why it was done.

How much work would it involve to just change the interface to take the request as an argument? Other providers would then just take the `code` form value?

Is it too much of a stretch to force v1 and v2 down the same interface?

I wouldn't consider these questions blockers, but I'd like to call it out.

auth/bitbucket/server/auth_config.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/concourse/atc"
+	"github.com/concourse/atc/auth/bitbucket"
+	"github.com/concourse/atc/auth/routes"
+	"github.com/hashicorp/go-multierror"
+	"github.com/tedsuo/rata"
+)
+
+type AuthConfig struct {
+	ConsumerKey string           `json:"consumer_key" long:"consumer-key" description:"Application consumer key for enabling Bitbucket OAuth"`
+	PrivateKey  privateKeyConfig `json:"private_key" long:"private-key" description:"Application private key for enabling Bitbucket OAuth, in base64 encoded DER format"`
```
vito: This should probably be a path to a file instead, similar to [other flags](https://github.com/concourse/atc/blob/750c42a630d8fc35dc58053993a53e0cbe8e2de9/atccmd/command.go#L104) of the same nature.

auth/bitbucket/cloud/auth_config.go
SHyx0rmZ: Yup. We we're still working out the differences between Bitbucket Cloud and Server. This should be functional before an eventual merge, but we wanted feedback on the rest of the pull request first.

auth/bitbucket/server/auth_config.go
SHyx0rmZ: Wouldn't this require the transfer of the file to the machine running ATC though?

auth/bitbucket/server/provider.go
SHyx0rmZ: So the other providers would take `code` from the `Request` themselves? That probably shouldn't be too hard. I'll give it a try.

We tried to avoid changing the interface, because we don't know how many people rely on it. A similar thing happened with the v1 vs. v2 routes.

Personally, I'd like to either change the v2 route so something like `/oauth/v2/` or pass information about the version through the Elm code, so we can decide which protocol to follow in the callback handler (and get rid of the second set of routes in the process).

auth/bitbucket/server/auth_config.go
vito: Depends on the automation tooling you've got. Our binaries already expect files for more complicated values like private keys. Our BOSH release just dumps templates into files on the machine, and our Docker instructions pass it all along via a volume mount.

I just think it's cleaner than having the key show up in the command's argument list; that could be revealed by `ps` and such (or `/proc/XXX/cmdline`). Granted, files on a disk aren't the greatest either, but at least you can have permissions configured appropriately.

auth/bitbucket/server/provider.go
vito: Yeah - let's switch it to `*http.Request`. All the things that rely on it should be within the ATC, so the scope is limited.

I think there's definitely something off about mixing v1 and v2 like this, but I think we should maybe pull this in and tackle that ourselves (unless you want to give it a go), with #202 in mind as well. With both of these merged, we'd end up with 4 auth flows:

* basic auth
* LDAP basic auth
* oAuth v2
* oAuth v1

We've generalized oAuth v2 with a common interface, but it doesn't seem like that was enough. We may need to make things even more dynamic so that even basic auth is no longer hardcoded.

auth/bitbucket/server/auth_config.go
SHyx0rmZ: I was under the impression that teams could modify their own auth, which would make putting the keys on the server problematic. But even if they could, the key would probably only have to be readable by `fly`, right?

auth/bitbucket/server/provider.go
SHyx0rmZ: I'll try to have a look at #202 and come up with something, but I won't be too sad if this gets pulled in after changing the interface, either.

auth/bitbucket/server/auth_config.go
tanzeeb: Yeah, it would only need to be readable by `fly`.

vito:
Thanks! Could you rebase this? Looks like the build's failing, and at least one of those compiler errors would be fixed by the changes we've made to the ATC PR task (which is in this repo, so until you rebase it won't run).

SHyx0rmZ:
Sure! I also hope to finish the support for projects and teams by the end of the week, so the rebase might take until then.

vito:
Sounds good, thanks!

SHyx0rmZ:
Hm, there's still one `panic` in `server.(*Provider).AuthCodeURL` that I'd like to get rid of. I think it would be best to change the interface here, too, so it returns `(string, error)`. Does that sound good?

vito:
@SHyx0rmZ Sure, sounds good.

SHyx0rmZ:
I think I'm not missing anything else, so I would now consider this "complete", unless you have any more feedback.

vito:
Looks good, thanks a bunch!

SHyx0rmZ:
Thanks for merging!

charnjit1234:
Is there any documentation for getting this setup?

PR #219, opened 2017-11-02
Add random assignment of tasks across eligible workers.

phillbaker:
This is an attempt to address https://github.com/concourse/concourse/issues/1741 (and maybe allow https://github.com/concourse/concourse/issues/675 to be addressed in the future).

It hides a new scheduling "algorithm" behind a new commandline flag which defaults to the current bucketing + random choice method. In the future this could be extended to to other algorithms like round robin or least resource utilization. All feedback welcome!

Added test coverage of the pool option, but not of the new ATC flag as the atc_command acceptance test only seemed to cover the non-trivial options.

cc @JohannesRudolph & thanks for pointing to the location of the scheduling code


jtarchie:
@phillbaker, thanks! Have you done any testing of this on a real life deployment? It would be interesting to see the metric data points from the before and after.

phillbaker:
@jtarchie thanks for taking a look. 

Our deployment unfortunately doesn't capture the logs/metrics 😒, but it _would_ be great to see a before/after of both job length times, resource utilization and job "gravity" over time. We're seeing issues build only after a couple of weeks, so I'm also afraid using our deployment would have a slow iteration cycle. 

Any suggestions/examples for gathering that kind of data efficiently?

atccmd/command.go
```diff
@@ -106,6 +106,7 @@ type ATCCommand struct {
 	ResourceCheckingInterval          time.Duration `long:"resource-checking-interval" default:"1m" description:"Interval on which to check for new versions of resources."`
 	OldResourceGracePeriod            time.Duration `long:"old-resource-grace-period" default:"5m" description:"How long to cache the result of a get step after a newer version of the resource is found."`
 	ResourceCacheCleanupInterval      time.Duration `long:"resource-cache-cleanup-interval" default:"30s" description:"Interval on which to cleanup old caches of resources."`
+	WorkerSchedulingMethod            string        `long:"worker-scheduling-method" default:"bucket" description:"How jobs shoudld be scheduled across the worker pool, options: 'bucket' or 'random'."`
```
vito: This could be configured as `choice:"bucket" choice:"random"` instead: https://godoc.org/github.com/jessevdk/go-flags

JohannesRudolph:
@vito's suggestions make sense from a code perspective. I was wondering if/how it would be possible to add better visualization of actual task distribution to the pipeline/UI. 

I know that `fly` reports the containers & worker assignment, but it would make issues a lot easier to debug if the UI could display on which worker a task is running (and by extension: total build time of a task would be great as well). Or maybe add a "workers" screen to the UI where one could in realtime see task distribution on workers (with running tasks and finished task containers visually separated)? 

All of these are entirely separate issues though

phillbaker:
> I have one idea to make the implementation better though: instead of injecting a bool, could we define a ContainerPlacementStrategy interface?

👍 

Took a stab at this, if this looks like a good start, I'll finish up moving over the tests as well.

phillbaker:
Updated as per your suggestion @vito, and finished moving the tests over. Let me know what you think!

atccmd/command.go
```diff
@@ -106,6 +106,7 @@ type ATCCommand struct {
 	ResourceCheckingInterval          time.Duration `long:"resource-checking-interval" default:"1m" description:"Interval on which to check for new versions of resources."`
 	OldResourceGracePeriod            time.Duration `long:"old-resource-grace-period" default:"5m" description:"How long to cache the result of a get step after a newer version of the resource is found."`
 	ResourceCacheCleanupInterval      time.Duration `long:"resource-cache-cleanup-interval" default:"30s" description:"Interval on which to cleanup old caches of resources."`
+	WorkerSchedulingMethod            string        `long:"worker-scheduling-method" default:"bucket" choice:"bucket" choice:"random" description:"How jobs shoudld be scheduled across the worker pool."`
```
vito: Could we rename this to `ContainerPlacementStrategy`, rename `bucket` to `volume-locality`, and have the description say "Method by which a worker is selected during container placement." or something?

I think that more precisely captures what it does.

phillbaker: Sounds good, updated!

topherbullock:
Thanks for the PR @phillbaker!

PR #221, opened 2017-11-14
[RFC] Add attempts to versioned_resources & endpoint

phillbaker:
This attempts to address point (2) from https://github.com/concourse/concourse/issues/413#issuecomment-233076618

> Retrying a build (this issue): I want to re-run the current build, either to see if it's flaky or to retry because something outside the build failed (e.g. github, a deployment, etc.).

It adds a counter (`attempts`) to the tuple that defines resource versions, so concourse would now control `resource_id`, `type`, and `attempts` and the resource's `check` provides the `version`. (I believe that the existing scheduler will work without changes, however, I had trouble following all of the logic in the scheduler package.) This could later be exposed both in fly (e.g. `fly -t ci reattempt PIPELINE/RESOURCE/ID`  and in the UI via a button similar to the "+" button on the resources page. 

Happy to receive feedback and comments - I believe this address some of the concerns raised in https://github.com/concourse/concourse/issues/413#issuecomment-311399474 about versions in the context of pipelines, but if this isn't the direction concourse is going (e.g. https://github.com/concourse/concourse/issues/1172) happy to close as well. Thanks!

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -180,6 +180,7 @@ func NewHandler(
 		atc.GetResourceVersion:            pipelineHandlerFactory.HandlerFor(versionServer.GetResourceVersion),
 		atc.EnableResourceVersion:         pipelineHandlerFactory.HandlerFor(versionServer.EnableResourceVersion),
 		atc.DisableResourceVersion:        pipelineHandlerFactory.HandlerFor(versionServer.DisableResourceVersion),
+		atc.ReattemptResourceVersion:      pipelineHandlerFactory.HandlerFor(versionServer.ReattemptResourceVersion),
 		atc.ListBuildsWithVersionAsInput:  pipelineHandlerFactory.HandlerFor(versionServer.ListBuildsWithVersionAsInput),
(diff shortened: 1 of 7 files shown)


vito:
Neat, thanks for taking a swing at this! It's a pretty hefty feature request.

I'm a little hesitant to start down this track with the [spatial resource flows](https://github.com/concourse/concourse/issues/1707) work in flight. I still think the comment you linked is an interesting vector to explore, but I suspect that in completing #1707 we'll have a deeper understanding of the problem space and be better suited to tackle the original problem.

The original problem may also end up being substantially mitigated when we're done with #1707. For example, it may be that in many cases where you need a re-trigger, in the new world it would be one row of a spatial pipeline branch, and "re-trigger" would actually just be "trigger".

I'm going to close this for now (sorry!), but I really do appreciate it. 😄

vito:
Closing per comment above. Thanks!

phillbaker:
Whoops, missed your earlier comment. Thanks for taking a look - makes sense. 👍

PR #222, opened 2017-11-21
[Work In Progress] Add Session Idle Timeouts

sharms:
@jmcarp and I did an initial implementation for idle timeouts and tested it with both 0 and 5m successfully using the current Concourse master branch.  Looking for feedback on if something like this would be the right approach, or if there are other ways we should investigate.

Thanks!


pivotal-issuemaster:
@sharms Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=222)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=222) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sharms Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=222)!

api/containerserver/hijack.go
```diff
@@ -192,6 +195,14 @@ func (s *Server) hijack(hLog lager.Logger, conn *websocket.Conn, request hijackR
 				}
 			} else {
 				_, _ = stdinW.Write(input.Stdin)
+				if s.interceptIdleTimeout != 0 {
+					idle.Reset(s.interceptIdleTimeout)
```
vito: This can probably just be right after the `case` line; it might as well apply to all inputs.

api/containerserver/hijack.go
```diff
@@ -192,6 +195,14 @@ func (s *Server) hijack(hLog lager.Logger, conn *websocket.Conn, request hijackR
...
+				}
+			}
+
+		case <-idle.C:
+			if s.interceptIdleTimeout != 0 {
+				errs <- fmt.Errorf("Idle timeout of %d was reached", s.interceptIdleTimeout)
```
vito: Could you define an error type for this instead? e.g.:

```
type InterceptTimeoutError struct {
    Duration time.Duration
}

func (err InterceptTimeoutError) Error() string {
    return fmt.Sprintf("idle timeout (%s) reached", err.Duration)
}
```

jmcarp:
Thanks for reviewing @vito ! Any tips on where tests should go? I don't see an obvious place to add tests in atc, but maybe this would fit in https://github.com/concourse/testflight/blob/master/flying/flying_test.go ?

vito:
@jmcarp I'm not sure about TestFlight since it's a deployment-wide configuration. A unit test could probably be added in the `api` package. I think it would only need to inject a low timeout value, initiate a websocket connection, and see that the connection breaks.

jmcarp:
@vito: I refactored a bit to make this easier to test. It's more verbose, but IMO allows for better tests. WDYT?

vito:
Looks good now! Is this ready for merge?

jmcarp:
Ready for merge as far as we're concerned--thanks for the feedback @vito !

PR #224, opened 2017-11-23
Use Kubernetes secrets as a credential manager

william-tran:
Address PR feedback in #189. 

- Lookup secrets with the same heuristic used in Vault.
- Use a prefix for namespaces to avoid collisions.

Testing locally:

Assuming you can use `kubectl` on your environment:

1. `kubectl create namespace concourse-main`
1. `kubectl config set-context $(kubectl config current-context) --namespace=concourse-main`
1. `kubectl create secret generic hello-world.secrets --from-literal "secret1=secretvalue1"`
1. edit concourse/dev/atc (or concourse/dev/atc-dockerdb) and add `--kubernetes-config-path=[userhome]/.kube/config` to the command line flags
1. `fly sp hello-world -c hello-world.yml` with the following pipeline

```
jobs:
- name: hello-world
  plan:
  - task: say-hello
    config:
      platform: linux
      image_resource:
        type: docker-image
        source: {repository: ubuntu}
      params:
        SECRET: ((secrets.secret1))
      run:
        path: echo
        args: ["$SECRET"]
```


william-tran:
Sorry for the false start... I'll re-open once we're ready.

PR #227, opened 2017-11-28
Allow tasks to be clickable behind legend

SwamWithTurtles:
Apologies if this is contributed incorrectly and/or contentious at all!

Our concourse installation has tasks to be laid out such that the legend sits over the tasks. Since this is a block element at the top, the tasks underneath it cannot be clicked on even if they are in progress.

image

Since the legend does not contain any links or clickable 'calls to action' we can use pointer-events as a quick fix to propagate the click events to the elements below (this works on all browsers except IE10 and below, which will act as before so there's no regression there.)

### DIFF
--- a/web/assets/css/production.less
+++ b/web/assets/css/production.less
@@ -292,6 +292,7 @@ li.prep-status {
 }
 
 .legend {
+  pointer-events: none;
   position: absolute;
   bottom: 1em;
   left: 1em;


pivotal-issuemaster:
@SwamWithTurtles Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=227)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=227) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@SwamWithTurtles Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=227)!

vito:
@jama-pivotal Seems simple - should we just merge this in?

jama-pivotal:
ermerghersh yes we should! sorry i was on vacation and just saw this

mhuangpivotal:
@SwamWithTurtles we made a manual commit (https://github.com/concourse/web/commit/698b46a2794e7a602d287c5c304fdcd793ef7521) because the web is now in a separate repo.

Thanks for the PR!

PR #228, opened 2017-11-29
Add a flag for using k8s secrets when running in cluster.

william-tran:
This makes using the Kubernetes credential manager an explicit choice when running inside Kubernetes, and also allows you to use a different credential manager when running in cluster.


william-tran:
ping @farcaller @vito. The helm chart recent added Vault support, so we can't force k8s secrets as the credential manager when running in cluster.

vito:
thanks!

PR #229, opened 2017-11-30
Support AWS system parameters for secret storage

surajbarkale:
This PR adds support for getting secrets from AWS SSM parameter storage. By default secrets are accessed by team & pipeline. However, this can be changed with command line parameter using templates.

Fixes: https://github.com/concourse/concourse/issues/1676


pivotal-issuemaster:
@surajbarkale Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=229)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=229) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@surajbarkale Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=229)!

pms1969:
@surajbarkale Legend.  Although I suspect you'll be asked for some tests.

surajbarkale:
@pms1969 I suspected as much :) I have been thinking about how to add tests since none of the other providers have them. Testing template logic is straightforward, however I am not sure about the `List()` variables function. Any feedback on parameter naming is appreciated since I have covered only the use case important to us.

pms1969:
> Any feedback on parameter naming is appreciated since I have covered only the use case important to us.

I think as long as there's a way to override the template that should be fine.  And as far as I can tell, it is possible (unless I've misunderstood).

I quite like the fact that you've allowed for both Instance Roles, and specified credentials.  Certainly saves someone having to ask for it.

vito:
Awesome, thanks for doing this!

Yeah, right now we're covering credential management at a higher-level, in [TopGun](https://github.com/concourse/topgun). This was done because unit tests at this level tend to be boring; the actual code is just glue using high-level libraries.

The `List` functionality is also currently not implemented or used by other managers, so that's fine. We might use it some day but for now we're punting.

One thing I'll block this on however is that we've stopped going with `vendor/` for pluggable interfaces like this. Each time it's been somewhat painful to maintain and we already have a method of tracking our dependencies in the [concourse/concourse repo](https://github.com/concourse/concourse). If you just strip all that out we can add the proper submodules when we merge this in. Just let us know of any special version constraints.

As for testing, I'm not sure what the best way to go forward is. If there are parts you suspect would be valuable to unit test, I'd obviously encourage it. I think we're missing a more general test harness for this, though; maybe something we can extract from TopGun.

Here are the tests for Vault: https://github.com/concourse/topgun/blob/master/vault_test.go

And here are the tests for CredHub: https://github.com/concourse/topgun/blob/master/credhub_test.go

We punted on k8s, though in hindsight we probably could have whipped up something that deployed k8s just so we can point to it for secret management.

Are there any gotchas with testing this? Would it be possible to add a similar suite to Vault and CredHub that's just configured via env vars and runs if they're configured?

I can provide assistance with running TopGun if so. Here's a quick tl;dr:

* Be on OS X or Linux.
* Clone https://github.com/concourse/concourse
* Run `./dev/vbox/setup`
* Run `./dev/vbox/topgun`

surajbarkale:
I have added some test cases to cover template parsing logic and I have removed the vendor folder. I am setting up Topgun and I shall ping you if I need any help.

creds/ssm/manager.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/concourse/atc/creds"
+)
+
+type SsmManager struct {
+	AwsAccessKeyID     string `long:"aws-access-key" description:"AWS Access key ID"`
+	AwsSecretAccessKey string `long:"aws-secret-key" description:"AWS Secret Access Key"`
+	AwsSessionToken    string `long:"aws-session-token" description:"AWS Session Token"`
+	AwsRegion          string `long:"aws-region" description:"AWS region to send requests to. Enviroment variable AWS_REGION is used if this flag is not provided."`
+	SecretTemplate     string `long:"secret-template" description:"AWS SSM parameter name template" default:"/{{.Team}}/{{.Pipeline}}/{{.Secret}}"`
```
itsdalmo: Regarding `"/{{.Team}}/{{.Pipeline}}/{{.Secret}}"`; as I commented in the issue, I was thinking that it might be nice to support a prefix (like `concourse/`), because without it we'd have to: 

- A: Give `/*` access to the ATC (not recommended).
- B: Change the policy every time you onboard a new team. 

However, in writing this comment I realise that this is not strictly needed for my use case (Concourse deployed using terraform) because I can just automate everything. But, perhaps it would still be nice to have for users that are doing e.g. Bosh deployments of Concourse?

creds/ssm/manager.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+	// Execute the template on dummy data to verify that it does not expect additional data
+	dummy := SsmSecret{Team: "team", Pipeline: "pipeline", Secret: "secret"}
+	err = secretTemplate.Execute(ioutil.Discard, &dummy)
+	if err != nil {
+		return err
+	}
+	// All of the AWS credential variables may be empty since credentials may be obtained via environemnt variables
+	// or other means. However, if one of them is propvided, then all of them must be provided.
```
itsdalmo: propvided => provided :D

creds/ssm/manager.go
itsdalmo: Since I haven't used parameter store before I did some sanity testing also, just throwing this in here in case it's useful in any way. Some rules:

- Parameter name cannot end with a `/`.
- Name cannot contain a `/` without also starting with a `/` (so `/some` and `/some/test` is allowed, `some/test` is not.)

I created 5 secrets:
- /some (just shows as 'some' in console)
- /some/test
- /some/test/value
- /other/secret

And tweaked the following policy:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ssm:DescribeParameters"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "ssm:*"
            ],
            "Resource": "<resource>"
        },
        {
            "Effect": "Allow",
            "Action": [
                "kms:Decrypt"
            ],
            "Resource": [
                "arn:aws:kms:eu-west-1:<account>:key/*"
            ]
        }
    ]
}
```

And `<resource>` replaced as follows:

- `arn:aws:ssm:eu-west-1:<account>:parameter*` - Access to everything.
- `arn:aws:ssm:eu-west-1:<account>:parameter/*` - Access to everything.
- `arn:aws:ssm:eu-west-1:<account>:parameter/some*` - Access to all `/some*`
- `arn:aws:ssm:eu-west-1:<account>:parameter/some/*` - Access to all `/some/*`, but not `/some`.

#### My takeaway

It seems SSM Parameter store is meant to be used with paths. Scoping IAM policies based on path seems to work as expected. I.e., I think the approach taken in this PR make sense :)

creds/ssm/ssm.go
```diff
@@ -0,0 +1,75 @@
...
+	err := s.secretTemplate.Execute(&buf, &SsmSecret{
+		Team:     s.teamName,
+		Pipeline: s.pipelineName,
+		Secret:   varName,
+	})
+	return buf.String(), err
+}
+
+func (s *Ssm) Get(varDef varTemplate.VariableDefinition) (interface{}, bool, error) {
+	secretName, err := s.buildSecretName(varDef.Name)
```
itsdalmo: Just noticed. It seems this implementation will only look at `/<team>/<pipeline>/<secret>` and not also attempt `/<team>/<secret>` if `<secret>` is not found for the specific pipeline. This contrasts with how the Vault backend  is implemented (see [here](https://github.com/concourse/atc/blob/master/creds/vault/vault.go#L30-L35)), and I think for consistency (and convenience) the SSM backend should work the same way.

surajbarkale: I shall add a `fallback-secret-template` that defaults to this path.

creds/ssm/manager.go
surajbarkale: I have added `concourse-` as prefix in the template. Since the path can have [upto 5 levels](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParametersByPath.html#EC2-GetParametersByPath-request-Path) only, it seemed prudent to provide more levels to the user. Policy setting should still work as expected since it does not care about the path.

itsdalmo: @surajbarkale 

I thought that if you reference a `((secret-value))` in a pipeline.yml, the current implementation of this (and other) secret backends will only check two paths:

- `/concourse/{{.Team}}/{{.Pipeline}}/secret-value`
- `/concourse/{{.Team}}/secret-value`

And that it's not possible to put e.g. `((nested/secret-value))` in pipeline.yml, and then retrieve `/concourse/{{.Team}}/{{.Pipeline}}/nested/secret-value`?

If I'm not mistaken about the above, I wonder if the users would actually be able to make use of the levels we save them by using `concourse-` as a prefix? (I.e., users would only run into trouble if they wanted to nest the concourse secrets under an existing path, which would require passing a new template via the command line anyway?)

creds/ssm/manager.go
```diff
@@ -0,0 +1,132 @@
...
+
+const DefaultSecretTemplate = "concourse-{{.Team}}/{{.Pipeline}}/{{.Secret}}"
+const DefaultFallbackTemplate = "concourse-{{.Team}}/{{.Secret}}"
+
+type SsmManager struct {
+	AwsAccessKeyID     string `long:"aws-access-key" description:"AWS Access key ID"`
+	AwsSecretAccessKey string `long:"aws-secret-key" description:"AWS Secret Access Key"`
+	AwsSessionToken    string `long:"aws-session-token" description:"AWS Session Token"`
+	AwsRegion          string `long:"aws-region" description:"AWS region to send requests to. Enviroment variable AWS_REGION is used if this flag is not provided."`
+	SecretTemplate     string `long:"secret-template" description:"AWS SSM parameter name template" default:"concourse-{{.Team}}/{{.Pipeline}}/{{.Secret}}"`
```
itsdalmo: I thought that if you reference a `((secret-value))` in a pipeline.yml, the current implementation of this (and other) secret backends will only check two paths:

- `/concourse/{{.Team}}/{{.Pipeline}}/secret-value`
- `/concourse/{{.Team}}/secret-value`

And that it's not possible to put e.g. `((nested/secret-value))` in pipeline.yml, and then retrieve `/concourse/{{.Team}}/{{.Pipeline}}/nested/secret-value`?

If I'm not mistaken about the above, I wonder if the users would actually be able to make use of the levels we save them by using `concourse-` as a prefix? (I.e., users would only run into trouble if they wanted to nest the concourse secrets under an existing path, which would require passing a new template via the command line anyway?)

creds/ssm/manager.go
surajbarkale: Ah, I initially thought that parameters allow `/` in names. Since that is not the case. I shall change to using 'concourse/' to match with existing [documentation](https://concourse.ci/creds.html#section_credential-lookup-rules).

creds/ssm/manager.go
surajbarkale: Thanks for all the feedback! I shall focus on topgun next.

surajbarkale:
@vito I have created [PR for topgun tests](https://github.com/concourse/topgun/pull/1) associated with this change. However, I am unable to successfully execute the tests on my machine. The SSM parameters are never passed from bosh to atc. I logged the commandline during test execution, and it results in following (with no ssm parameters):
```
/var/vcap/packages/atc/bin/atc --bind-ip 0.0.0.0 --bind-port 8080 --peer-url http://10.234.0.2:8080 \
--external-url http://10.234.0.2:8080 --postgres-host 10.234.0.2 --postgres-port 5432 \
--postgres-database atc --postgres-sslmode disable --postgres-user atc \
--postgres-password dummy-password --postgres-connect-timeout 5m \
--no-really-i-dont-want-any-auth --log-level debug --basic-auth-username  \
--basic-auth-password  --auth-duration 24h --github-auth-client-id  --github-auth-client-secret  \
--uaa-auth-client-id  --uaa-auth-client-secret  --uaa-auth-auth-url  --uaa-auth-token-url \
--uaa-auth-cf-url  --uaa-auth-cf-ca-cert /var/vcap/jobs/atc/config/cf_ca_cert \
--generic-oauth-client-id  --generic-oauth-client-secret  --generic-oauth-auth-url  \
--generic-oauth-scope  --generic-oauth-token-url  --generic-oauth-display-name  \
--session-signing-key /var/vcap/jobs/atc/config/token_signing_key --resource-checking-interval 1m \
--old-resource-grace-period 5m --resource-cache-cleanup-interval 30s \
--container-placement-strategy volume-locality --gc-interval 30s --build-tracker-interval 10s \
--baggageclaim-response-header-timeout 1m --cli-artifacts-dir /var/vcap/packages/fly \
--yeller-api-key  --yeller-environment  --influxdb-database  --influxdb-username  \
--influxdb-password  --riemann-host  --riemann-port 5555 --riemann-service-prefix  \
--metrics-host-name concourse-0 --metrics-attribute bosh-deployment:concourse-topgun-1 \
--metrics-attribute bosh-job:concourse --x-frame-options
```

BTW I had to also execute `scripts/sync-package-specs` before `dev/vbox/setup` to be able to get tests to compile. 

Since I am not at all familiar with bosh, your help would be appreciated to make these tests work.

vito:
@surajbarkale Awesome!

To pass the properties along to BOSH and then to the ATC, you'll need to configure them as properties in the [job spec](https://github.com/concourse/concourse/blob/master/jobs/atc/spec), and use them in the [`atc_ctl.erb` template](https://github.com/concourse/concourse/blob/master/jobs/atc/templates/atc_ctl.erb). You can have a look at how it's done for Vault as an example.

That would then be another thing you could PR (to https://github.com/concourse/concourse).

surajbarkale:
Finally the topgun tests have passed! @vito can you please review all the linked PRs and manage the merge into concourse?

vito:
@surajbarkale Great! I'll check it out now.

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+
+const DefaultPipeSecretTemplate = "/concourse/{{.Team}}/{{.Pipeline}}/{{.Secret}}"
+const DefaultTeamSecretTemplate = "/concourse/{{.Team}}/{{.Secret}}"
+
+type SsmManager struct {
+	AwsAccessKeyID     string `long:"aws-access-key" description:"AWS Access key ID"`
+	AwsSecretAccessKey string `long:"aws-secret-key" description:"AWS Secret Access Key"`
+	AwsSessionToken    string `long:"aws-session-token" description:"AWS Session Token"`
+	AwsRegion          string `long:"aws-region" description:"AWS region to send requests to. Enviroment variable AWS_REGION is used if this flag is not provided."`
+	PipeSecretTemplate string `long:"pipe-secret-template" description:"AWS SSM parameter name template used for pipeline specific paramter" default:"/concourse/{{.Team}}/{{.Pipeline}}/{{.Secret}}"`
```
vito: Let's just spell this out as "pipeline".

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+	TeamSecretTemplate string `long:"team-secret-template" description:"AWS SSM parameter name template used for team specific paramter" default:"/concourse/{{.Team}}/{{.Secret}}"`
```
vito: Typo: "paramter"

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+const DefaultPipeSecretTemplate = "/concourse/{{.Team}}/{{.Pipeline}}/{{.Secret}}"
+const DefaultTeamSecretTemplate = "/concourse/{{.Team}}/{{.Secret}}"
+
+type SsmManager struct {
+	AwsAccessKeyID     string `long:"aws-access-key" description:"AWS Access key ID"`
+	AwsSecretAccessKey string `long:"aws-secret-key" description:"AWS Secret Access Key"`
+	AwsSessionToken    string `long:"aws-session-token" description:"AWS Session Token"`
+	AwsRegion          string `long:"aws-region" description:"AWS region to send requests to. Enviroment variable AWS_REGION is used if this flag is not provided."`
```
vito: This can be done as `env:"AWS_REGION"` in the struct tags. (You can then remove that blurb from the description; I think it'll show up in the help-text automatically.)

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, err
+	}
+	if parse.IsEmptyTree(t.Root) {
+		return nil, errors.New("secret template should not be empty")
+	}
+	return t, nil
+}
+
+func (manager SsmManager) IsConfigured() bool {
+	return manager.AwsRegion != "" || os.Getenv("AWS_REGION") != ""
```
vito: Getenv not needed if struct tag mentioned above is used.

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if manager.AwsSessionToken == "" {
+		return errors.New("must provide aws session token")
+	}
+
+	return nil
+}
+
+func (manager SsmManager) NewVariablesFactory(log lager.Logger) (creds.VariablesFactory, error) {
+	log.Info("Creating new SSM variables factory", lager.Data{
```
vito: Logs should be in the form of `creating-new-ssm-variables-factory`.

Though, these log lines should probably be removed before merging, or at least moved down to `Debug`.

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+		"pipe-secret-template": manager.PipeSecretTemplate,
+		"team-secret-template": manager.TeamSecretTemplate,
+	})
+	config := &aws.Config{}
+	if manager.AwsRegion != "" {
+		config.Region = &manager.AwsRegion
+	}
+	if manager.AwsAccessKeyID != "" {
+		log.Info("Using AWS credentials provided by user", lager.Data{"aws-access-key": manager.AwsAccessKeyID})
```
vito: This log should definitely be removed or at least not emit the key ID.

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+		config.Credentials = credentials.NewStaticCredentials(manager.AwsAccessKeyID, manager.AwsSecretAccessKey, manager.AwsSessionToken)
+	}
+
+	session, err := session.NewSession(config)
+	if err != nil {
+		log.Error("Failed to establish AWS session", err)
```
vito: Should be in the form of `failed-to-construct-aws-session`. (Sorry, I don't think we have this convention on contributor docs anywhere.)

creds/ssm/manager_factory.go
```diff
@@ -0,0 +1,26 @@
...
+	creds.Register("ssm", NewSsmManagerFactory())
+}
+
+func NewSsmManagerFactory() creds.ManagerFactory {
+	return &ssmManagerFactory{}
+}
+
+func (factory *ssmManagerFactory) AddConfig(group *flags.Group) creds.Manager {
+	manager := &SsmManager{}
+	subGroup, err := group.AddGroup("SSM Credential Management", "", manager)
```
vito: Can we call this `AWS SSM Credential Management`?

creds/ssm/manager_factory.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		panic(err)
+	}
+	subGroup.Namespace = "ssm"
```
vito: ...And make this namespace `aws-ssm`?

creds/ssm/ssm.go
```diff
@@ -0,0 +1,113 @@
...
+		Pipeline: s.PipelineName,
+		Secret:   varName,
+	})
+	return buf.String(), err
+}
+
+func (s *Ssm) Get(varDef varTemplate.VariableDefinition) (interface{}, bool, error) {
+	for _, st := range s.SecretTemplates {
+		if secret, err := s.buildSecretName(st, varDef.Name); err != nil {
+			s.log.Error("Failed to build SSM parameter path from secret", err, lager.Data{
```
vito: Should be `failed-to-build-ssm-parameter-path-from-secret`.

creds/ssm/ssm.go
```diff
@@ -0,0 +1,113 @@
...
+				"template": st.Name(),
+				"secret":   varDef.Name,
+			})
+			return nil, false, err
+		} else if value, found, err := s.getParameterByName(secret); err != nil {
+			s.log.Error("Failed to get SSM paramter by name", err, lager.Data{
```
vito: Should be `failed-to-get-ssm-parameter-by-name`.

creds/ssm/ssm.go
```diff
@@ -0,0 +1,113 @@
...
+				"template": st.Name(),
+				"secret":   secret,
+			})
+			return nil, false, err
+		} else if found {
+			return value, true, nil
+		} else if value, found, err = s.getParameterByPath(secret); err != nil {
+			s.log.Error("Failed to get SSM paramter by path", err, lager.Data{
```
vito: Should be `failed-to-get-ssm-parameter-by-path`.

creds/ssm/ssm.go
```diff
@@ -0,0 +1,113 @@
...
+				"secret":   varDef.Name,
+			})
+			return nil, false, err
+		} else if value, found, err := s.getParameterByName(secret); err != nil {
+			s.log.Error("Failed to get SSM paramter by name", err, lager.Data{
+				"template": st.Name(),
+				"secret":   secret,
+			})
+			return nil, false, err
+		} else if found {
```
vito: This chaining of `else-if`s feels a bit unconventional. I would just move the `.getXXX` calls out of the `if`-block, since it returns other important values aside from the `err`:

```go
value, found, err = s.getXXX();
if err != nil {
  // ...
}

if found {
  return value, true, nil
}

value, found, err = s.getXXX()
// ...
```

creds/ssm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/concourse/atc/creds"
+)
+
+const DefaultPipeSecretTemplate = "/concourse/{{.Team}}/{{.Pipeline}}/{{.Secret}}"
+const DefaultTeamSecretTemplate = "/concourse/{{.Team}}/{{.Secret}}"
+
+type SsmManager struct {
+	AwsAccessKeyID     string `long:"aws-access-key" description:"AWS Access key ID"`
```
vito: These `aws-` prefixes can probably be removed along with the namespace change to `aws-ssm` mentioned (far) below.

surajbarkale:
During development I found it surprising that a secret `obj.field` expands to parameter `team/pipeline/obj/field`. I am not sure if it is communicated clearly in documentation or if I am too thick to understand it.

This may trip up other users as well which will add SSM parameter `team/pipeline/obj.field` and get confused.

vito:
Thanks for following up, the changes look good now. :) I'm not sure how to follow up on your last comment there, but I'll at least approve the current code.

surajbarkale:
@vito I have uncovered a problem with empty pipeline name and added tests to cover it. My comment was a hint to use when updating the [documentation](https://concourse.ci/creds.html). I apologize that I am unable to do so on my own.

vito:
Test failure with those latest changes - that test can probably just be removed.

https://ci.concourse.ci/builds/731652#L5a3817aa:54:63

vito:
Thanks a bunch! Merging this in now.

PR #231, opened 2017-12-06
[WIP] Add arbitrary labels to prometheus metrics.

jmcarp:
Not sure if this is the most elegant implementation, but here goes. I want to be able to attach arbitrary labels to my prometheus metrics, like cluster name (we run multiple concourses, because reasons). I know we already have `Metrics.Attributes`, but we don't currently pass those attributes through to prometheus metrics, because we're whitelisting a set of attributes to expose for each metric--which is a good thing, since we want to avoid high-cardinality labels like build id. This patch adds a new flag for labels to be added as constants to all prometheus metrics.

cc @sharms @TimSimmons

### DIFF
--- a/metric/emitter/prometheus.go
+++ b/metric/emitter/prometheus.go
@@ -30,8 +30,9 @@ type PrometheusEmitter struct {
 }
 
 type PrometheusConfig struct {
-	BindIP   string `long:"prometheus-bind-ip" description:"IP to listen on to expose Prometheus metrics."`
-	BindPort string `long:"prometheus-bind-port" description:"Port to listen on to expose Prometheus metrics."`
+	BindIP   string            `long:"prometheus-bind-ip" description:"IP to listen on to expose Prometheus metrics."`
+	BindPort string            `long:"prometheus-bind-port" description:"Port to listen on to expose Prometheus metrics."`
+	Labels   map[string]string `long:"prometheus-labels" description:"Labels to add to all Prometheus metrics. Can be specified multiple times." value:name:"NAME:VALUE"`
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


TimSimmons:
So this is generally kind of an [anti-pattern](https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels): `If you ever find yourself wanting to apply the same label to all of your metrics, stop.`

It's recommended to take care of labels like these using Prometheus configuration.
It's trivial if you're using [`file_sd`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config) or [`static_config`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#static_config).

Otherwise, it's probably a use case for a [`relabel_config`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config). The docs there are kind of scary, but usually the information is available already via the `instance` name, or perhaps from another tag in whichever service discovery method being used.

As an example, I often want the region as a label for my metrics.
```
- job_name: prod-app
  file_sd_configs:
  - files:
    - /opt/prometheus/services/prod-app.json
  relabel_configs:
  - regex: '[^\.]+\.([^\.]+)\..*'
    replacement: ${1}
    source_labels:
    - __address__
    target_label: region
```

Using the automatic `__address__` label before it disappears, I can append a `region` label.

I'd recommend against implementing something like this here.

jmcarp:
Interesting. For my use case, I'm using service discovery from https://github.com/bosh-prometheus/bosh_exporter, which AFAIK doesn't give me a way to figure out what I'm looking for (deployment name) from the labels it provides. I'll check with the maintainer and see if we can solve the problem from the service discovery side instead of the exporter.

jmcarp:
Fixed with a patch to bosh_exporter at https://github.com/bosh-prometheus/bosh_exporter/pull/15.

PR #232, opened 2017-12-07
Create migration for new BTREE index on check_order column of version…

christophermancini:
…_resources table.

This PR replaced https://github.com/concourse/atc/pull/218

### DIFF
--- a/db/migration/bindata.go
+++ b/db/migration/bindata.go
@@ -4,6 +4,8 @@
 // db/migration/migrations/1510262030_initial_schema.up.sql
 // db/migration/migrations/1510670987_update_unique_constraint_for_resource_caches.down.sql
 // db/migration/migrations/1510670987_update_unique_constraint_for_resource_caches.up.sql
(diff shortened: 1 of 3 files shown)


vito:
Looks good, just leaving a comment to kick the CI build and give it another shot. Looks like a flake.

vito:
Dang, looks like I forgot to merge this after kicking the build. Would you mind rebasing to resolve the bindata conflict?

(Note to self: bindata is probably gonna be a pain point for PRs going forward - maybe we should investigate building it in the pipeline instead.)

mhuangpivotal:
Hi @christophermancini, I took the liberty to merge this manually. Thanks!

vito:
We had to roll this back as it caused significant performance issues on our large-scale environment (Wings). We're not 100% sure why, but things perform better without this index. Pretty surprising!

I'm going to capture an optimization for this in concourse/rfcs#1 under the "Implementation Notes" section at the bottom. It'll be even more important once we have the new interface, as it'll be checking across multiple spaces and with many more versions in the history (all of'em).

PR #233, opened 2017-12-07
Fix k8s secret resolution logic

farcaller:
Apparently there's an issue with the current code. If concourse only has access to "$SECRET" secret and not "$PIPELINE.$SECRET", then apiserver will return a 403 which fails fast, not allowing concourse to try the second option.

This PR makes sure that concourse will try both options before failing, also improves logging.

### DIFF
--- a/creds/kubernetes/kubernetes.go
+++ b/creds/kubernetes/kubernetes.go
@@ -25,18 +25,11 @@ func (k Kubernetes) Get(varDef template.VariableDefinition) (interface{}, bool, 
 
 	secret, found, err := k.findSecret(namespace, pipelineSecretName)
 
-	if !found && err == nil {
-		secret, found, err = k.findSecret(namespace, secretName)
+	if found {
+		return k.getValueFromSecret(secret)
 	}
 
-	if err != nil {
-		k.logger.Error("k8s-secret-error", err, lager.Data{
-			"namespace":          namespace,
-			"pipelineSecretName": pipelineSecretName,
-			"secretName":         secretName,
-		})
-		return nil, false, err
-	}
+	secret, found, err = k.findSecret(namespace, secretName)
 
 	if found {
 		return k.getValueFromSecret(secret)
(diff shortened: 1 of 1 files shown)


creds/kubernetes/kubernetes.go
```diff
@@ -23,22 +23,15 @@ func (k Kubernetes) Get(varDef template.VariableDefinition) (interface{}, bool,
 	var pipelineSecretName = k.PipelineName + "." + varDef.Name
 	var secretName = varDef.Name
 
-	secret, found, err := k.findSecret(namespace, pipelineSecretName)
```
vito: I preferred the code before; `foo, found, err` is more explicit. I think we should explicitly check for `403` instead.

By not checking the type of error, this could result in interesting behavior where e.g. both `$PIPELINE.$SECRET` and `$SECRET` are defined, but the network flakes out for the first query, and works for the second, resulting in a query occasionally returning the wrong secret.

william-tran:
@farcaller How do you set up rbac so that the client only has access to one secret in the namespace and not the other?

creds/kubernetes/kubernetes.go
farcaller: Sorry, I missed the comment. Will follow up with a better code.

farcaller:
@william-tran 

```yaml
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: Role
metadata:
  name: concourse
rules:
  - apiGroups: ['']
    resourceNames:
      - containers.mysecret # only needed as a workaround for this bug
      - mysecret # actual secret
    resources: [secrets]
    verbs: [get]
```

william-tran:
@farcaller I assumed the typical use case would be a ClusterRoleBinding that allows all secrets in the team specific namespace to be read by concourse, where the namespace is set up for the sole purpose of providing secret values to all pipelines in a particular team. See my comment here: https://github.com/concourse/atc/pull/189#issuecomment-341502356. You can control what secrets concourse has access to by just adding and removing secrets in that namespace. One downside to this approach is either the duplication of secret values used outside concourse, or needing to provide specific bindings to each non-concourse secret usage. 

I think the situation you're solving for is one where:
- You have an existing namespace that contains secrets
- Your team name and namespace prefix are set up to map to this namespace
- Not all secrets in this namespace should be made available to concourse.

Are we on the same page here?

vito:
This feels related to https://github.com/concourse/atc/pull/254?

vito:
Sorry but I'm gonna close this for now since I'm not sure it's going anywhere. There are other PRs open related to k8s secrets and I'd like to minimize the mental gymnastics of correlating them.

Thanks for kicking this off though!

PR #234, opened 2017-12-08
Add a basic metric emitter for the newrelic insights api

novas0x2a:
This adds an emitter for the [New Relic Insights API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/insert-custom-events-insights-api). This actually works as-is, but New Relic charges on event volume, with a total account max on event types, so I was wondering if you had guidance on ways to cut down on the event volume without losing too much useful information.

Are the metrics on the [dashboard](https://metrics.concourse.ci/) the only ones you consider crucial? I could add a --newrelic-high-priority-only or something to filter on that set.

### DIFF
--- a/metric/emitter/newrelic.go
+++ b/metric/emitter/newrelic.go
@@ -1,0 +1,87 @@
+package emitter
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"code.cloudfoundry.org/lager"
+	"github.com/concourse/atc/metric"
+)
+
+type NewRelicEmitter struct {
+	client *http.Client
+	url    string
+	apikey string
+	prefix string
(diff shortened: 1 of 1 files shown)


metric/emitter/newrelic.go
```diff
@@ -0,0 +1,85 @@
...
+
+	req, err := http.NewRequest("POST", emitter.url, bytes.NewBuffer(payloadJSON))
+	if err != nil {
+		logger.Error("failed-to-construct-request", err)
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("X-Insert-Key", emitter.apikey)
+
+	_, err = emitter.client.Do(req)
```
vito: This should close the response body to prevent a leak.

vito:
thanks!

PR #236, opened 2017-12-14
Add Vault Cache and Exponential Backoff for Login

rfliam:
Add a cache to the Vault credentials manager. The cache is flagable
and defaults to off. The cache uses the vault set lease duration and
re-fetches secrets halfway to the expiration time.

Add an exponential backoff for login and renewal of vault credentials.
This slows down spam to vault in the event of a bad configuration or
during heavy load. This backoff starts at 1 second and max's out at 64
seconds. This is currently not configurable.

As part of the change refactor so to test both the cache and the retry
strategies. Add unit tests to this effect. It also adds more comments
in the code and cleanups up some go naming stuff.

This change has _not_ been tested with a full deployment yet (I think the  PR will do that for me? ;p)

This is needed because if a user configures secrets in to be passed as params to a check resource it
is trivial to inadvertently attack your own secrets infrastructure. 

I considered trying to add a "meta" credentials cache, but I'm not sure that a single caching strategy/manager will be able to correctly abstract all possible credential stores expected caching behaviors. For example k8s already has a  [sophisticated caching mechanism](https://godoc.org/k8s.io/client-go/tools/cache) available in its standard library.

### DIFF
--- a/creds/vault/api_client.go
+++ b/creds/vault/api_client.go
@@ -1,0 +1,111 @@
+package vault
+
+import (
+	"path"
+	"sync/atomic"
+	"time"
+
+	"code.cloudfoundry.org/lager"
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+// The APIClient is a SecretReader which maintains an authorized
+// client using the Login and Renew functions.
+type APIClient struct {
+	clientValue *atomic.Value
+	logger      lager.Logger
+	config      AuthConfig
+}
+
+// NewAPIClient with the associated authorization config and underlying vault client.
+func NewAPIClient(logger lager.Logger, client *vaultapi.Client, config AuthConfig) *APIClient {
+	ac := &APIClient{
+		clientValue: &atomic.Value{},
+		logger:      logger,
+		config:      config,
+	}
+	ac.setClient(client)
+	return ac
+}
+
+// Read must be called after a successful login has occurred or an
+// un-authorized client will be used.
+func (ac *APIClient) Read(path string) (*vaultapi.Secret, error) {
+	return ac.client().Logical().Read(path)
+}
+
+func (ac *APIClient) loginParams() map[string]interface{} {
+	loginParams := make(map[string]interface{})
+	for _, param := range ac.config.Params {
+		loginParams[param.Name] = param.Value
+	}
+
+	return loginParams
+}
+
+// Login the APIClient using the credentials passed at
+// construction. Returns a duration after which renew must be called.
+func (ac *APIClient) Login() (time.Duration, error) {
+	// If we are configured with a client token return right away
+	// and trigger a renewal.
+	if ac.config.ClientToken != "" {
+		ac.setClient(clientWithToken(ac.client(), ac.config.ClientToken))
+		return 0, nil
+	}
+
+	logger := ac.logger.Session("login")
+	client := ac.client()
+	secret, err := client.Logical().Write(path.Join("auth", ac.config.Backend, "login"), ac.loginParams())
+	if err != nil {
+		logger.Error("failed", err)
+		return time.Second, err
+	}
+	logger.Info("succeeded", lager.Data{
+		"token-accessor": secret.Auth.Accessor,
+		"lease-duration": secret.Auth.LeaseDuration,
+		"policies":       secret.Auth.Policies,
+	})
+
+	ac.setClient(clientWithToken(client, secret.Auth.ClientToken))
+	return time.Duration(secret.Auth.LeaseDuration) * time.Second, nil
+}
+
+// Renew the APIClient login using the credentials passed at
+// construction. Must be called after a successful login. Returns a
+// duration after which renew must be called again.
+func (ac *APIClient) Renew() (time.Duration, error) {
+	logger := ac.logger.Session("renew")
+	client := ac.client()
+	secret, err := client.Auth().Token().RenewSelf(0)
+	if err != nil {
+		logger.Error("failed", err)
+		return time.Second, err
+	}
+	logger.Info("succeeded", lager.Data{
+		"token-accessor": secret.Auth.Accessor,
+		"lease-duration": secret.Auth.LeaseDuration,
+		"policies":       secret.Auth.Policies,
+	})
+
+	ac.setClient(clientWithToken(client, secret.Auth.ClientToken))
+	return time.Duration(secret.Auth.LeaseDuration) * time.Second, nil
+}
+
+func (ac *APIClient) client() *vaultapi.Client {
+	return ac.clientValue.Load().(*vaultapi.Client)
+}
+
+func (ac *APIClient) setClient(client *vaultapi.Client) {
+	ac.clientValue.Store(client)
+}
+
+func clientWithToken(client *vaultapi.Client, token string) *vaultapi.Client {
+	// TODO: This is dangerous, we are relying on the vault
+	// client being ok with a shallow copy....
+	// but...
+	// the old code worked ;p
+	clientCopy := *client
+	newClient := &clientCopy
+	newClient.SetToken(token)
+	return newClient
+}
--- a/creds/vault/cache.go
+++ b/creds/vault/cache.go
@@ -22,17 +22,19 @@ type Cache struct {
 	newItems chan time.Time
 	sr       SecretReader
 	context  context.Context
+	maxLease time.Duration
 }
 
 // TODO: Should a cache have a max size to
 // prevent unbounded growth?
 
 // NewCache using the underlying vault client.
-func NewCache(sr SecretReader) *Cache {
+func NewCache(sr SecretReader, maxLease time.Duration) *Cache {
 	c := &Cache{
 		cache:    make(map[string]*cachedSecret),
 		newItems: make(chan time.Time, 100),
 		sr:       sr,
+		maxLease: maxLease,
 	}
 	go c.reaperThread()
 	return c
@@ -101,6 +103,9 @@ func (c *Cache) Read(path string) (*vaultapi.Secret, error) {
 	// This is a problem in any implementation, as lease duration could be 1s and the
 	// build will _probably_ take longer  than that.
 	dur := time.Duration(secret.LeaseDuration) * time.Second / 2
+	if c.maxLease != 0 && dur > c.maxLease {
+		dur = c.maxLease
+	}
 
 	// Store the secret in cache
 	cs = &cachedSecret{
--- a/creds/vault/cache.go
+++ b/creds/vault/cache.go
@@ -1,0 +1,118 @@
+package vault
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+type cachedSecret struct {
+	deadline time.Time
+	secret   *vaultapi.Secret
+}
+
+// A Cache caches secrets read from a SecretReader until the lease on
+// the secret expires. Once expired the credential is proactively
+// deleted from cache to maintain a smaller cache footprint.
+type Cache struct {
+	sync.RWMutex
+	cache    map[string]*cachedSecret
+	newItems chan time.Time
+	sr       SecretReader
+	context  context.Context
+}
+
+// TODO: Should a cache have a max size to
+// prevent unbounded growth?
+
+// NewCache using the underlying vault client.
+func NewCache(sr SecretReader) *Cache {
+	c := &Cache{
+		cache:    make(map[string]*cachedSecret),
+		newItems: make(chan time.Time, 100),
+		sr:       sr,
+	}
+	go c.reaperThread()
+	return c
+}
+
+func (c *Cache) reapCache() time.Time {
+	c.Lock()
+	defer c.Unlock()
+
+	var smallestNext time.Time
+	for k, secret := range c.cache {
+		if time.Now().After(secret.deadline) {
+			delete(c.cache, k)
+			continue
+		}
+		if secret.deadline.Before(smallestNext) {
+			smallestNext = secret.deadline
+		}
+	}
+	return smallestNext
+}
+
+func (c *Cache) reaperThread() {
+	sleep := time.NewTimer(1 * time.Second)
+	defer sleep.Stop()
+	var nextWakeup time.Time
+	for {
+		select {
+		case <-sleep.C:
+			nextWakeup = c.reapCache()
+			if !nextWakeup.IsZero() {
+				sleep.Reset(nextWakeup.Sub(time.Now()))
+			}
+		case t := <-c.newItems:
+			if t.Before(nextWakeup) {
+				continue
+			}
+			nextWakeup = t
+			sleep.Reset(t.Sub(time.Now()))
+		}
+	}
+}
+
+// Read a secret from the cache or the underlying client if not
+// present.
+func (c *Cache) Read(path string) (*vaultapi.Secret, error) {
+	// If we have the secret in our cache just return it
+	c.RLock() // don't use defer because we want to agressively release this lock
+	cs, cached := c.cache[path]
+	c.RUnlock()
+
+	if cached && time.Now().Before(cs.deadline) {
+		return cs.secret, nil
+	}
+
+	// Otherwise fetch the secret using the client. Clients are
+	// thread safe for read use.
+	secret, err := c.sr.Read(path)
+	if err != nil || secret == nil {
+		return secret, err
+	}
+
+	// We will renew the item in half the lease duration to resolve an inherent race
+	// in this setup: What if the secret becomes invalid _during_ the build. We don't
+	// want to be issuing secrets that expire in (fex) 100ms.
+	// This is a problem in any implementation, as lease duration could be 1s and the
+	// build will _probably_ take longer  than that.
+	dur := time.Duration(secret.LeaseDuration) * time.Second / 2
+
+	// Store the secret in cache
+	cs = &cachedSecret{
+		deadline: time.Now().Add(dur),
+		secret:   secret,
+	}
+	c.Lock()
+	c.cache[path] = cs
+	c.Unlock()
+
+	// Tell the reaper thread it has new items to cleanup
+	c.newItems <- cs.deadline
+
+	return secret, nil
+}
--- a/creds/vault/manager.go
+++ b/creds/vault/manager.go
@@ -36,10 +36,10 @@ type VaultManager struct {
 type AuthConfig struct {
 	ClientToken string `long:"client-token" description:"Client token for accessing secrets within the Vault server."`
 
-	Backend       string        `long:"auth-backend" description:"Auth backend to use for logging in to Vault."`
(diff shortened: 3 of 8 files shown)


pivotal-issuemaster:
@rfliam Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=236)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=236) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@rfliam Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=236)!

jtarchie:
Adding caching at the application level may be masking the issue. It seems your `vault` deployment might need to be scaled out to handle the traffic coming into it. Though concourse is causing the load, delegating a caching layer just for `vault` doesn't seem right to me.

What have you tried actually scaling out `vault`?

rfliam:
Our vault instance could handle the load, it is just completely unnecessary. The vault API is _designed_ to use caching. That is why [lease_duration](https://www.vaultproject.io/docs/concepts/lease.html) exists. Most [vault clients](https://www.npmjs.com/package/vault-client) feature it.

Concourse is doing the "wrong thing" by not caching a secret that is _intended_ to be cached.

vito:
/cc @eugenMayer - this PR would probably make you happy

also ref. https://github.com/hashicorp/vault/issues/3651

Kind of a bummer that there's no blessed Go client that implements this kind of functionality; I was hesitant to implement it ourselves because caching credentials sounds like a bad idea if there's no precedent or strong need for it. But it sounds like there's a means for it via Vault's API, and precedent in other popular libraries, though it's still kind of a bummer that we have to implement the caching ourselves.

Given that, I want to be super careful about this and make sure we're taking into consideration any "gotchas" with the implementation (e.g. accidentally using an expired credential after it's been rotated, etc.), and any security risks involved (e.g. holding all the secrets in the ATC's head). I say this having not yet reviewed the code in detail.

Are there other Go consumers of Vault we can look to as an example? Have other projects run into this pain point?

rfliam:
Agreed on the go client not having caching being a bit of a bummer. But on the other hand the nature of the cache is _probably_ use case specific anyways. Concourse for example is a read mostly cache. Requests for previously un-requested secrets are going to be _very_ low.

Hence having a cache focused on fast reads and low memory overhead for concourse makes sense. We could use [golang-lru](https://github.com/hashicorp/golang-lru) from hashicorp but it doesn't actively purge records and does cache eviction on the main thread.

The cache I implemented is very efficient for this use case, only performing maintenance when keys expire. However it does not have a strict upper bounds in memory size. This is a complicated matter for vault secrets as they are of arbitrary size so having a max number of secrets only helps some in preventing large memory footprints.

It's also worth mentioning it looks like the original code re-implemented [renewer](https://github.com/hashicorp/vault/blob/master/api/renewer.go) and I followed suite. So that should probably be changed out.

creds/vault/cache.go
```diff
@@ -0,0 +1,116 @@
...
+// deleted from cache to maintain a smaller cache footprint.
+type Cache struct {
+	sync.RWMutex
+	cache    map[string]*cachedSecret
+	newItems chan time.Time
+	sr       SecretReader
+}
+
+// TODO: Should a cache have a max size to
+// prevent unbounded growth?
```
EugenMayer: if we do that, does it then rather make sense going for redis in a term, that this is already offered there and its also a memory cache and flushed when we restart the services ( security reasons ).

creds/vault/manager.go
```diff
@@ -17,6 +18,8 @@ type VaultManager struct {
 
 	PathPrefix string `long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
 
+	Cache bool `bool:"cache" default:"false" description:"Cache returned secrets for their lease duration in memory"`
```
EugenMayer: Should we also add a adjustable value for secret.LeaseDuration - so people can adjust to their usecase? E.g in our case, those secrets are 99% read only and if we write, we would restart ATC. People mighr use vault differently with AWS and rotation, so setting this down a lot ( or disable the cache ). But being able to define the general aggressiveness would be a way to help people adjusting to their use-case very easy, with very little effort on our side. What do you think?

jefferai:
Hi there,

A long while back Vault did in fact issue leases for the `generic` (now `kv`) backend but we took that out because it was causing problems with exactly the kind of behavior that Concourse displays -- refreshing from each client once a second was a really great way to generate a massive number of leases and bring Vault crashing down. We kept the ability for the writer to round-trip a lease duration, with the idea that the writer often knows how often they're going to write new values, so it can use that value as an expected lifetime hint for readers.

What we usually see is that people will set a value for the lease duration that is significantly less than the actual expiration, but large enough to not cause high load. For example, let's say you rotate a credential every 12 hours and keep the old one active for 2 hours to allow clients to adjust. The writer of the credential may set the lease duration to 90 minutes, which should be enough time for all clients to retrieve the new value (with 5399 less reads from Vault per client in the interim compared to checking every second :-D ).

Although the API could do local caching, you'd need to know how long the cache should hold that value anyways, and still be capable of retrieving new values when outside the caching period. So a built-in cache doesn't buy you much other than the ability to continue checking every second instead of simply setting your timer to check at the interval given by the secret.

Eventually we hope to add an eventing API, but it's not going to appear in a "soon" timeframe. If your underlying data store supports an eventing API you could do some mapping of Vault storage to data store storage and use an event on the underlying data store path to trigger a re-read from Vault clients, but I would generally avoid that unless absolutely necessary for your use-case.

vito:
@jefferai Thanks for helping us figure this out!

I think there's one thing to clarify: Concourse doesn't ever explicitly refresh anything or do anything special with leases, let alone refresh once a second. We fetch a credential any time we need it, and throw it away when we're done with it.

Here's what's happening on Concourse's side:

1. Say a user has configured a `resource` in their Concourse pipeline (resources are just YAML config representing an integration point, i.e. a Git repo or an S3 bucket). Say this resource has a configuration pointing to `((private_key))`.
1. Concourse periodically performs a `check` operation on every pipeline resource, by default once per minute. This is to discover things like new commits to said Git repo.
1. If a resource has a `((parameter))`, every time we `check` we'll resolve those parameters by fetching them from Vault.
1. We run the resource's `check` script, passing it the resource's configuration, with the parameterized credentials filled with the values fetched from Vault.

After running the `check` script, we don't do anything with the credentials - in fact, they're out of the ATC's head as soon as we've started running the script. Once the checking interval fires again, we'll just fetch the credentials anew. If they're rotated, we'll have the new values. Or, a user may have reconfigured the pipeline in the meantime, and we may be fetching a new set of credentials. Who knows!

So from Concourse's perspective, these credentials are short-lived. They aren't used throughout the duration of the Concourse server running, for example. The stack is stateless, so we don't have any 'refreshing' mechanism, we just don't use the credentials for very long in the first place.

Considering that, and given the existence of lease durations as a way for the credential writer to convey a safe validity window (separate from expiry), this feels like a generic client-side cache. I'm just worried about us being the ones to write it. ;)

...And quite worried that this may result in a single process that contains literally all of the credentials for all of the pipelines and all of the teams in its head. Not as a question of memory usage, but as a question of risk, and to what extent that defeats the point of using Vault in the first place.

jefferai:
Hi @vito 

Thanks for the explanation, that helps me understand what Concourse is doing here.

Not going in order:

> ...And quite worried that this may result in a single process that contains literally all of the credentials for all of the pipelines and all of the teams in its head. Not as a question of memory usage, but as a question of risk, and to what extent that defeats the point of using Vault in the first place.

I can't really help you assess whether that risk is acceptable or not, but it's certainly a better stance than persisting to disk. Using Vault helps sort out the "not persisting to disk in plaintext" part of the equation (and a lot of other functionality), so even if you are caching things on the client side you're still getting benefit.

> The stack is stateless, so we don't have any 'refreshing' mechanism, we just don't use the credentials for very long in the first place.

Between this and the other discussion of client side caching, I'm a bit confused -- do you keep your API clients around? If so that's not really stateless, but if not, I don't really see how caching in the clients will help.

vito:
@jefferai 

> I can't really help you assess whether that risk is acceptable or not, but it's certainly a better stance than persisting to disk. Using Vault helps sort out the "not persisting to disk in plaintext" part of the equation (and a lot of other functionality), so even if you are caching things on the client side you're still getting benefit.

We're still getting a lot of benefit from Vault in plenty of ways, I'm just wary of any step back we may make in security. CI systems are an especially lucrative attack target. Either way it's on us to characterize the risk involved and figure out what we can/should do. 

> Between this and the other discussion of client side caching, I'm a bit confused -- do you keep your API clients around? If so that's not really stateless, but if not, I don't really see how caching in the clients will help.

Sorry, all I meant by stateless was that the credentials themselves aren't e.g. stashed in an object somewhere in need of periodic refreshing. They're fetched and then thrown away. Though I guess that's how a lot of Vault consumers work anyway. We do probably have a single API client, so caching would be fine.

jefferai:
Hi @vito,

Thanks for the explanation. If you do work on a caching system I'd love for it to be in the API client since I think others could benefit. I have a couple thoughts around it:

* One way might be to allow the Renewer function in the API to work on secrets that don't have a lease ID by providing a function that can be run to generate a new Secret. That way the renewer can look at the lease duration and know that after e.g. 95% of the duration has expired, run the function to fetch a new one. (It would also allow the renewer to fetch new secrets generally, rather than only renew them.)

* Much of Consul-Template got pulled into a library a while back, which is how it's embedded into Nomad. Based on a template it knows how to fetch (and appropriately handle refetching) for a large variety of Vault secrets. The downside is that templates have to be written to use it.

Probably the first option would be the way to go for now. I do think there should be controls around disabling caching, perhaps on a per-secret (or per-path?) basis, in case the duration that comes back can't be trusted (or some secret shouldn't be kept around in memory).

rfliam:
# Some Design Thoughts:

## Current Security Model and Risks

Currently concourse keeps secrets for all team accessible by a single login managed in the ATC. If an attacker achieves some sort of "escalation" he can pull that token, clieky key, and client cert and access all secrets for _all_ teams. While this information is not kept in memory, the token is, and it would still be readily available.

Escalation for this token is somewhat difficult as it requires attacking the ATC. The "exploit vector" here is largely through the ATC API. The most common methodology here would be to dump the heap or the configuration of the ATC.

Exploits that escalate using containers in pipeline steps are prevented from attacking the ATC if it is run on physically isolated nodes (this is probably a best practice for the future). However, privilege escalations on a worker could be used to attack _other_ pipelines secrets that happen to be local. If the ATC is local this could also be attacked.

## The Impact of Caching

Maintaining secrets in memory offers an increased risk in "timeliness" of attack. If the client certificate is ip addressed pinned or the ATC but not the secrets back end is accessible this voids some of its benefits. With a cache a user must _only_ dump the heap to access the secrets as opposed to dumping the heap repeatedly, as secrets are in flight.

In practice this additional risk is low as most heap dumps are the result of existing remote code execution which can be crafted to attack these secrets directly.

## Possible Changes

### Per-Pipeline Secrets Management

Maintaining a per team or per pipeline login to vault could help reduce the magnitude of any breach. The secrets configuration could be set on the _pipeline_ rather then globally for the ATC. The trade off is this information would have to be persisted. However, the secrets back-end could be used for this. The ATC would be configured with a secrets back-end in which to store and access the pipeline specific secrets config. 

###  Changing Where and How Secrets Access Works

Another increase in security would be from "splitting" the secrets access and decryption. In this model the access to the secret would reside with the ATC or Pipeline, but the secret itself would be further [encrypted with the private key of the worker]. The ATC would access the secret and forward it the worker which would then decrypt the secret for use. This adds some security as it requires a working exploit on both the worker _and_ the atc.

### Add Build Signing 

One thing I would _love_ to see is the addition of a build singing infrastructure. In this infrastructure _tasks_ could request that concourse sign a resource with it's private keys so _traceable_ builds could be used. Another usage would be requiring deployments to come from the build infrastructure itself, signing deployment requests with the build infrastructures private keys.

[1] More correctly the secret would be stored with a randomly generated AES encryption key, and that key would be encrypted with the workers public key.

# Proposals
## Add the Cache (in the Vault Code)
Compared to today's model the cache is only a minor increase in risk. However, it is nearly _required_ for most secrets backends. As evidenced by #233 and [vault#3651](https://github.com/hashicorp/vault/issues/3651) secrets without caching can cause significant problems.

I very much like jefferai's suggestion to disable caching on a per path basis.

## Future
Look towards per-pipeline secrets and build signing first IMO. But that's up to you guys.

jefferai:
A couple of comments:

> The trade off is this information would have to be persisted. However, the secrets back-end could be used for this. The ATC would be configured with a secrets back-end in which to store and access the pipeline specific secrets config.

You may want to think about using `transit` for this -- either directly for encryption/decryption or to allow dissemination of an encrypted data key that a worker can decrypt via transit and hold locally in memory.

> One thing I would love to see is the addition of a build singing infrastructure. In this infrastructure tasks could request that concourse sign a resource with it's private keys so traceable builds could be used. Another usage would be requiring deployments to come from the build infrastructure itself, signing deployment requests with the build infrastructures private keys.

Just in case you haven't seen it, `transit` also has three different key types that can be used for signing (rsa, ed25519, ecdsa-p256) :-)

jefferai:
BTW, I just mention the signing since the text discusses workers' private keys; I'm not sure how those are disseminated/stored, but if they already have a Vault token, then you could just keep the keys in Vault and guard access to it.

rfliam:
Another thought here:

The cache can use [memguard](https://github.com/awnumar/memguard) to help prevent buffer overflow attacks.

jefferai:
Be careful with that line of thinking -- Go already prevents such attacks (subject to things working the way they're supposed to) by taking memory management out of your hands. memguard needs to advertise buffer overflow protection because it works around Go's memory management, which opens it up to a host of potential attacks that it then needs to mitigate.

I've been watching memguard for a while, and while I think it's an interesting library, as far as I can tell the main use of memguard is if you are concerned about values living for long periods in memory after you've given them up and don't want to try mitigating by periodically triggering garbage collections in Go. It won't stop someone that has root from getting your memory -- it will just make it harder via obfuscation.

Unless you feel super strongly about needing memguard, I'd just use something that supports expiry and/or running functions on eviction like https://github.com/patrickmn/go-cache or similar.

william-tran:
>I considered trying to add a "meta" credentials cache, but I'm not sure that a single caching strategy/manager will be able to correctly abstract all possible credential stores expected caching behaviors. For example k8s already has a sophisticated caching mechanism available in its standard library.

I'm a k8s user. I'd be happy with a simple caching strategy that applied to all credential managers, with a configurable TTL defaulting to 60 seconds. k8s secrets don't convey anything around expiry, and giving watch or list permissions to implement anything fancier for caching should be avoided:

https://kubernetes.io/docs/concepts/configuration/secret/#best-practices
> For these reasons watch and list requests for secrets within a namespace are extremely powerful capabilities and should be avoided, since listing secrets allows the clients to inspect the values if all secrets are in that namespace. The ability to watch and list all secrets in a cluster should be reserved for only the most privileged, system-level components.

There is also a PR to fix the k8s secret lookup so that access to all secrets in a namespace isn't assumed: https://github.com/concourse/atc/pull/233.

creds/vault/cache.go
sampeinado: ghfhj

vito:
Just chiming in to say we haven't forgotten about this, we're just working on getting 3.9 out the door before merging this in as it's already pretty huge. Thanks for your patience!

creds/vault/reauther.go
```diff
@@ -0,0 +1,73 @@
...
+			continue
+		}
+		ra.success()
+		close(ra.LoggedIn)
+		time.Sleep(delay)
+		break
+	}
+
+	// TODO: What if that token becomes invalid (was deleted for
+	// example)? Shouldn't we try to login again...?
```
vito: Yep! I think this is how we can consolidate with #256.

creds/vault/cache.go
vito: Hmm, yeah this could be really interesting for large multi-tenant Concourse installations. Some credentials can be fairly large (e.g. private keys), so it could add up quickly. You could always just not use caching, but chances are you probably *want* caching in that same scenario, so it becomes a question of which you sacrifice: ATC RAM vs. Vault CPU.

I think it's fine to ship without this, but maybe the thing to do here is just have a configurable max retention for each cache? This could either be time-based or usage based (i.e. each time a credential's cache is hit, we keep it for a bit longer, still limited by its expiry obviously).

creds/vault/manager.go
```diff
@@ -83,5 +86,14 @@ func (manager VaultManager) NewVariablesFactory(logger lager.Logger) (creds.Vari
 		return nil, err
 	}
 
-	return NewVaultFactory(logger, client, manager.Auth, manager.PathPrefix), nil
+	c := NewAPIClient(logger, client, manager.Auth)
+
+	// TODO: Configurable?
```
vito: Yeah, might as well make these configurable. They'd be right at home next to the auth max duration flag from #256.

Also, why 64 seconds? :P

rfliam:
Addressed comments from code reviews. Added that flagable max lease duration for the cache per @EugenMayer .

rfliam:
Added.

vito:
happening

vito:
Seems like this is causing 3 TopGun tests to fail consistently:

* https://ci.concourse-ci.org/teams/main/pipelines/main/jobs/topgun/builds/1187
* https://ci.concourse-ci.org/teams/main/pipelines/main/jobs/topgun/builds/1188
* https://ci.concourse-ci.org/teams/main/pipelines/main/jobs/topgun/builds/1190
* https://ci.concourse-ci.org/teams/main/pipelines/main/jobs/topgun/builds/1191

Will take a closer look.

vito:
Found a couple problems.

First problem: code was blocking on `<-loggedIn` in the variables factory, because this channel would only  be written to upon logging in. I've changed this to a channel that's closed the first time we successfully log in, so these channel reads will never block again.

Second problem: when copying the Vault client and setting a token, I observed a deadlock in the stack that happened because copying the client also copies its mutexes (they're non-pointer values, relying on the struct), and it happened to copy it while the mutex was locked. I'm gonna find a better way to do this since it already had a TODO with a warning on it. The warning was right!

jefferai:
Use https://godoc.org/github.com/hashicorp/vault/api#Client.Clone instead of copying a client.

creds/vault/api_client.go
```diff
@@ -0,0 +1,111 @@
...
+	return ac.clientValue.Load().(*vaultapi.Client)
+}
+
+func (ac *APIClient) setClient(client *vaultapi.Client) {
+	ac.clientValue.Store(client)
+}
+
+func clientWithToken(client *vaultapi.Client, token string) *vaultapi.Client {
+	// TODO: This is dangerous, we are relying on the vault
+	// client being ok with a shallow copy....
```
jefferai: It isn't :-)

jefferai:
Can I ask -- is there a reason you wrote ReAuther instead of using https://godoc.org/github.com/hashicorp/vault/api#Renewer ?

PR #238, opened 2017-12-16
Increase the throttling of the k8s client.

william-tran:
Default values of 5 qps and 10 burst are too low, set it to 100 requests
per second. Fixes #1893.


vito:
Seems reasonable even in light of the caching discussion (which at present would only improve the story for Vault). Thanks!

PR #241, opened 2018-01-15
remove timeout in db AfterSuite

rosenhouse:
postgres can be slow to terminate.

I'd be happy to raise the timeout to 20 or 60 seconds, but @vito [suggested](https://concourseci.slack.com/archives/C1X5A4AJX/p1516055728000020?thread_ts=1516054819.000319&cid=C1X5A4AJX) we just remove it completely.


vito:
thanks!

PR #239, opened 2018-01-04
Modify query to improve performance while maintaining functionality

christophermancini:
We find this query in our slow query logs continuously and this change should cut its execution time by about 60%.

Thanks @bhcleek for your help.

PR #243, opened 2018-01-18
2nd attempt to improve scheduler query

christophermancini:
After #239 failed to scale linearly for pipelines with significant amounts of versioned resource history, this is another attempt at improving performance of this query. In this approach, I am fetching the last 10,000 versioned_resource_ids into a csv string, then passing them to the scheduler query, removing the 3 resource and versioned_resources table joins. This should cause the scheduler query to become very cheap and may perform even better than my previous submission.


christophermancini:
I may have shaved 30s off the DB test suite... :)

vito:
I'm wondering if this could just be simplified even further. All the query is trying to do is find the latest modified timestamp of all versioned resources, build inputs, and build outputs, for comparing against the ATC's cached timestamp.

Could we just store a timestamp somewhere and increment it whenever we write to either of those tables? At that point does it even need to be a number? Could it just be a sequence that we increment?

christophermancini:
@vito if I am understanding correctly, create a table, `pipeline_cache`, that stores `pipeline_id`, `cache_key`. Create an insert / update trigger to update the record in this table every time build_outputs, build_inputs, or versioned_resources gets changed.

If so, I can update this PR to support that.

vito:
We haven't used triggers for anything else yet and I've got a healthy fear of the unknown, but I'd be down to review a triggers-based implementation.

I would recommend calling the table something more specific, i.e. `pipeline_versions_db_index` or something.

vito:
I'm gonna close this for now since it's getting a bit stale and the next approach may end up being different anyway. Thanks!

PR #246, opened 2018-02-07
Detect cycles between jobs due to passed constraints.

jmcarp:
If a pipeline has a dependency cycle between jobs due to passed constraints (for example, job b depends on job a, job c depends on job b, and job a depends on job c), then the configuration is invalid. Also, the browser hits an infinite loop trying to render it. This patch detects `passed` cycles and errors on validation if any cycles are found.

Figured this out with @wjwoodson.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -195,6 +195,7 @@ func validateJobs(c Config) ([]Warning, error) {
 	warnings := []Warning{}
 
 	names := map[string]int{}
+	graph := NewGraph()
 
 	for i, job := range c.Jobs {
 		var identifier string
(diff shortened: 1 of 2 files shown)


jmcarp:
If `passed` cycles are actually useful, and if they work aside from the UI, then maybe we should fix the UI instead of making this change. I don't know when I would use them, and I find them a little hard to reason about, but that's just me.

I'm happy to switch this to a warning, or file a feature request for the UI to support cycles (I don't know elm well enough to write that patch yet).

vito:
@jmcarp Actually, we happen to be working on a new pipeline UI already as part of the spaces work (concourse/concourse#1707). I suspect the new UI will work fine, since we're removing the complicated line-drawing logic, so I'm going to just close this so we're not spending effort on the old UI.

Thanks again!

PR #247, opened 2018-02-07
Allow ephemeral workers

tanner-bruce:
I'm creating this initial commit for ephemeral workers to get feedback on my direction and pointers on things I may have missed


pivotal-issuemaster:
@tanner-bruce Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=247)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=247) to manually synchronize the status of this Pull Request.

timrchavez:
Hi :wave:. Could please provide clarification around what problem this work is trying to solve (for people looking from the outside in) or link this to a specific issue? This _sounds_ like something I should care about, but more context would help.

api/workers_test.go
```diff
@@ -755,6 +760,7 @@ var _ = Describe("Workers API", func() {
 			fakeWorker.TagsReturns([]string{"some-tag"})
 			fakeWorker.StateReturns(db.WorkerStateRunning)
 			fakeWorker.TeamNameReturns("some-team")
+			fakeWorker.EphemeralReturns(false)
```
vito: Might be better to have this return `true`, since it'll also be false in the response if we forget to set the field in the API.

db/worker_lifecycle.go
```diff
@@ -24,6 +25,26 @@ func NewWorkerLifecycle(conn Conn) WorkerLifecycle {
 	}
 }
 
+func (lifecycle *workerLifecycle) DeleteUnresponsiveEphemeralWorkers() ([]string, error) {
+	query, args, err := psql.Delete("workers").
+		Where(sq.Eq{"ephemeral": true}).
+		Where(sq.Expr("expires < NOW()")).
+		Where(sq.Eq{"state": string(WorkerStateRunning)}).
```
vito: Hmm could this result in a situation where you've started landing/retiring an ephemeral worker but then it never goes away when it's unresponsive?

Should we just remove the state check?

gc/workercollector.go
```diff
@@ -26,7 +26,15 @@ func (wc *workerCollector) Run() error {
 	logger.Debug("start")
 	defer logger.Debug("done")
 
-	affected, err := wc.workerLifecycle.StallUnresponsiveWorkers()
+	affected, err := wc.workerLifecycle.DeleteUnresponsiveEphemeralWorkers()
+
```
vito: small nit: no extra space here; `err` assignment is always flush with its `err != nil` check

vito:
@tanner-bruce Sorry it took us so long to initially review this. Just checking in to see if you still have time to iterate on this. If not we're happy to pick things up from here!

vito:
This doesn't seem to work with the binary distribution.

It looks like we share the config struct that provides the flags between `concourse/worker` and `concourse/bin`, but then we construct an `atc.Worker` independently in both, and `concourse/bin` wasn't updated to grab the `Ephemeral` flag. Can we consolidate this code?:

* https://github.com/concourse/worker/blob/344d6196df9323f36f027fb3dc86dc97ff3a1582/start/start_runner.go#L75
* https://github.com/concourse/bin/blob/6ef6b355774caf40d940c260b3f287b3af78a6d1/cmd/concourse/worker_linux.go#L74
* https://github.com/concourse/bin/blob/6ef6b355774caf40d940c260b3f287b3af78a6d1/cmd/concourse/houdini.go#L36

PR #248, opened 2018-02-09
Throw jsonapi 400 instead of empty 500 on resource type not found.

jmcarp:
Running `check-resource` on a resource with an undefined resource type throws a 500 with an empty response body. This patch throws a 404 with a jsonapi-formatted payload instead. This should make for clearer error messages, and also fixes part of https://github.com/concourse/concourse/issues/1443.

Note: ideally a resource with a missing resource type would fail validation. I'm guessing that hasn't happened because some resource types are globally available.

cc @wjwoodson

### DIFF
--- a/api/resources_test.go
+++ b/api/resources_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/google/jsonapi"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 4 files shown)


jmcarp:
How about `500`? Not sure any other `5xx` codes would fit. On the other hand, this should only happen when the user does something wrong, like misspelling a resource type--maybe `400` would make sense.

Separately, it looks like `go-concourse` parses errors differently by status code. Errors with code `400` are assumed to have a specific non-jsonapi structure, and other error codes don't get parsed as json at all: https://github.com/concourse/go-concourse/blob/master/concourse/check_resource.go#L50-L60. If concourse is moving to jsonapi, at least for errors, it might make sense to check for `Content-Type: application/vnd.api+json` instead.

jmcarp:
Ping @vito: should I just flip this to a `500`?

vito:
Woops, sorry, missed this.

I think 400 could make sense actually since it seems more likely that this
would happen due to a typo. Even if it's not a typo and you're missing
workers, it could make sense from the perspective of only resource types
provided by workers (or custom types) are "valid". Ignoring the fact that
we don't perform this validation when the pipeline is submitted, and a
worker dropping out could make it invalid.

To be honest I think we could get away with either. They can both be
correct based on different interpretations. Up to you! My issue was mainly
that 404 in particular felt ambiguous.

jmcarp:
Thanks, updated!

vito:
thanks!

PR #250, opened 2018-02-14
improve the new relic metrics emitter

novas0x2a:
- consolidate the container and volume periodic metrics
- omit some debugging metrics
- fix silent dropping of metrics with "(" in them
- add a catch-all metric for alerting metrics

### DIFF
--- a/metric/emitter/newrelic.go
+++ b/metric/emitter/newrelic.go
@@ -12,18 +12,30 @@ import (
 	"github.com/concourse/atc/metric"
 )
 
-type NewRelicEmitter struct {
-	client *http.Client
-	url    string
-	apikey string
-	prefix string
-}
+type (
+	stats struct {
+		created interface{}
+		deleted interface{}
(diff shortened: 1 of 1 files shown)


metric/emitter/newrelic.go
```diff
@@ -85,3 +108,82 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
 	resp.Body.Close()
 }
+
+func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
+	payload := make(fullPayload, 0)
+
+	switch event.Name {
+
+	// These are the simple ones that only need a small name transformation
+	case "build started":
```
vito: small nit, but rather than `fallthrough` this could just list the values in one `case`. i always forget that `fallthrough` even exists :P

vito:
thanks!

PR #253, opened 2018-03-04
WIP: Begin adding rebuild support

evanphx:
This attempts to address https://github.com/concourse/concourse/issues/413 by add a mechanism to create a new build using the input resources of an existing one.

This does not deal with the issue of versioning any output resources as that situation is quite thorny and best addressed via a different route.

This branch does not yet run because I haven't sorted out how to get the migrations working with needing to update the views for a new column on builds.

### DIFF
--- a/api/buildserver/restart.go
+++ b/api/buildserver/restart.go
@@ -1,0 +1,92 @@
+package buildserver
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/concourse/atc/api/present"
+	"github.com/concourse/atc/db"
+)
+
+type rebuildJob struct {
+	db.Job
+	oldBuild db.Build
+}
+
+func (r rebuildJob) CreateBuild() (db.Build, error) {
+	return r.CreateRebuild(r.oldBuild)
+}
+
+func (s *Server) RebuildBuild(build db.Build) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		logger := s.logger.Session("restart-build")
+
+		pipeline, found, err := build.Pipeline()
+		if err != nil {
+			logger.Error("failed-resolve-pipeline", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if !found {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		job, found, err := pipeline.Job(build.JobName())
+		if err != nil {
+			logger.Error("failed-resolve-job", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if !found {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		if job.Config().DisableManualTrigger {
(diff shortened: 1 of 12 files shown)


vito:
Awesome!

Gonna pull in @Lindsayauchin early here since there will most definitely be UI implications. This older comment may be useful context for how we reasoned about this functionality before: https://github.com/concourse/concourse/issues/413#issuecomment-233076618

The main thing I get hung up on is where the new build should show up, and what number should it have. The simplest approach would be to just give it whatever the next build number is, but would it be confusing to have the "latest" build (say 135) actually be a re-trigger of an older build (say, 113)? Should it be shown as `113 #2` or something?

Scheduling wise everything would work fine at least, since build order doesn't matter (only resource version ordering).

> This does not deal with the issue of versioning any output resources as that situation is quite thorny and best addressed via a different route.

This should be fine since `put`s are idempotent, and it should just record the same versions as an output of a new build. Because scheduling is based on resource ordering, which comes from `check`, this shouldn't have any negative impact on the rest of the pipeline, either.

> This branch does not yet run because I haven't sorted out how to get the migrations working with needing to update the views for a new column on builds.

Yeah, those materialized views are super annoying. :/ I wish migrations could be applied to them directly or something... We should keep an eye on the pain level here.

vito:
Chatted w/ @Lindsayauchin and @jama-pivotal in person. I think it may make the most sense to have a re-triggered build effectively replace or inject itself just after the targeted build in the history, rather than appear as a new build.

The thinking here is that if I have build history (G = green, R = red):

```
5 G, 4 G, 3 R, 2 G, 1 G
```

And I re-trigger `3 R`, and it goes red, that shouldn't cause the pipeline to show the job as "red". The mental model for the job state in the pipeline is "latest build", i.e. what's going to be propagating through the pipline, so I don't think re-triggers should become the "latest".

It should still be easy to tell that a build was re-triggered, so maybe we could show the history as something like this:

```
5 G, 4 G, 3-2 R, 2 G, 1 G
```

This is also semantically closer to how the scheduling works, as mentioned before, since the recency of a build run has no impact on how the jobs are scheduled - re-triggering an older build would just mark those older versions as outputs when it succeeds, which could allow them to be candidates for subsequent jobs, but it won't cause them to run if there's more recent stuff that already satisfies them.

There should still be some way to view older runs of re-triggered builds, though. Not sure how those would be reached - probably some UI cue when viewing a re-triggered build?

@evanphx Does this feel intuitive?

evanphx:
Certainly having the rebuild replace the old one in the history, visually, does make a lot of sense. Putting the rebuild at the end with a new build number is super confusing because it causes folks to be confused why a build went back in time and grabbed older versions of resources.

timrchavez:
I got really excited when I saw this work get picked up, but it's been over a month without an update. Wondering if I could help in anyway?

vito:
I'm gonna go ahead and close this for now since it's growing stale, though I would like to capture the discussion we had here somewhere. @timrchavez mentioned picking this back up in Discord, maybe we can just reference this PR once it's revived?

Thanks again @evanphx for kicking this off! Hope you find your way back to Concourse some time. :slightly_smiling_face:

timrchavez:
Yah we have someone (@nelsam) starting to take a look at resuming the work on this. Referencing the PR sounds good to me.

PR #254, opened 2018-03-13
Readded default namespace as well as functinality to look for multipl…

pontusarfwedson:
Adds functionality to:
* Be able to have secrets in a namespace named after our liking (by passing the variable `default-namespace` it will then look for secrets in that namespace, instead of forcing the user to have a namespace called `namespacePrefix+teamName`.
* Be able to have all of our credentials stored in one collective secret, or/and one collective secret per pipeline. So by passing in the variable `secrets-name` it will then look for our credentials within that k8s secret as well as within `secrets-name+pipelineName`. This allows us to create one k8s secret which includes all our encrypted credentials instead of having 1 secret for every credential.
* If no `default-namespace` is given, then it will look within `namespacePrefix+teamName` like previously, hence that functionality is not removed.
* If no `secrets-name` is given it will not look for any credentials in such a collective secret, hence the previous functionality still exists if a user elects to implement it that way. I considered removing the previous functionality completely but felt that more flexibility is positive in this case. 
* It will search in the order: pipeline.credName[value] -> credName[value] -> secretsName[credName] -> secretsName-pipeline[credName]

I believe this PR adds more flexibility to the user instead of making k8s secrets function exactly like Vault just because Vault is designed a certain way. Also, not boxing in the user to have to keep specific naming conventions for the namespace. And we should be clear about what a secret is. In K8S a secret is not a single K/V. One secret can contain multiple fields. And we should take advantage of that.

### DIFF
--- a/creds/kubernetes/kubernetes.go
+++ b/creds/kubernetes/kubernetes.go
@@ -11,29 +11,43 @@ import (
 )
 
 type Kubernetes struct {
-	Clientset       *kubernetes.Clientset
-	TeamName        string
-	PipelineName    string
-	NamespacePrefix string
-	logger          lager.Logger
+	Clientset        *kubernetes.Clientset
+	TeamName         string
+	PipelineName     string
+	NamespacePrefix  string
+	DefaultNamespace string
+	SecretsName      string
+	logger           lager.Logger
 }
 
 func (k Kubernetes) Get(varDef template.VariableDefinition) (interface{}, bool, error) {
 	var namespace = k.NamespacePrefix + k.TeamName
-	var pipelineSecretName = k.PipelineName + "." + varDef.Name
-	var secretName = varDef.Name
+
+	if k.DefaultNamespace != "" {
+		namespace = k.DefaultNamespace
+	}
+	var pipelineCredentialName = k.PipelineName + "." + varDef.Name
+	var credentialName = varDef.Name
 
 	secret, found, err := k.findSecret(namespace, pipelineSecretName)
 
 	if !found && err == nil {
-		secret, found, err = k.findSecret(namespace, secretName)
+		secret, found, err = k.findSecret(namespace, credentialName)
+	}
+
+	if !found && err == nil && k.SecretsName != "" {
+		secret, foundCollective, err = k.findSecret(namespace, k.SecretsName)
+	}
+
+	if !found && !foundCollective && err == nil && k.SecretsName != "" {
+		secret, foundCollective, err = k.findSecret(namespace, k.SecretsName+"-"+k.PipelineName)
 	}
 
 	if err != nil {
 		k.logger.Error("k8s-secret-error", err, lager.Data{
 			"namespace":          namespace,
 			"pipelineSecretName": pipelineSecretName,
-			"secretName":         secretName,
+			"credentialName":     credentialName,
 		})
 		return nil, false, err
 	}
@@ -42,10 +56,15 @@ func (k Kubernetes) Get(varDef template.VariableDefinition) (interface{}, bool, 
 		return k.getValueFromSecret(secret)
 	}
 
+	if foundCollective {
+		return k.getCredentialValueFromCollectiveSecret(credentialName, secret)
+	}
+
 	k.logger.Info("k8s-secret-not-found", lager.Data{
 		"namespace":          namespace,
 		"pipelineSecretName": pipelineSecretName,
-		"secretName":         secretName,
+		"credentialName":     credentialName,
+		"secretsName":        k.SecretsName,
 	})
 	return nil, false, nil
 }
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@pontusarfwedson Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=254)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=254) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@pontusarfwedson Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=254)!

vito:
@pontusarfwedson Sorry for the delay, thanks for submitting this! I'm not very familiar with the k8s secrets backend, so my only concern with pulling in changes is whether it risks exposing team A's credentials to team B, e.g. by specifying an absolute path. This is why the Vault backend is so limited at the moment.

@william-tran Would you mind taking a look at this when you get time?

william-tran:
@pontusarfwedson 

>Be able to have all of our credentials stored in one collective secret, or/and one collective secret per pipeline.
>This allows us to create one k8s secret which includes all our encrypted credentials instead of having 1 secret for every credential.

Doesn't a reference of `((common-secrets.api-key))` achieve the same effect, other than the need to specify the secret name? See 
https://github.com/kubernetes/charts/tree/master/stable/concourse#kubernetes-secrets

pontusarfwedson:
Hey @william-tran 
I actually missed the information regarding `common-secrets`. Is that a secret name that is statically set? We cannot name it what we want somehow? Browsing the secrets in k8s might get a little confusing with one simply being named `common-secrets`.  Apart from that, that is actually the functionality I was going for. I am not sure where that functionality is implemented though. Looking at https://github.com/concourse/atc/blob/master/creds/kubernetes/kubernetes.go I get the impression it only looks for the key `value` in `pipeline`+`secret` or just in `secret`. 

Regarding your last part. My aim here was to keep all the existing functionality (like the one I just described) but add the possibility of all credentials in one secret, very similar to the functionality described in the docs you sent. And also make the choice of namespace more flexible. For example, running your common services in the `default` namespaces is very common for k8s where Concourse defaults their team to `main`. If one runs the setup, the one cannot put the desired secrets in the default namespace. Also, if we run our services in the namespace `team1` and then have a `team1` in concourse then we cannot make our secrets go to that namespace because the current implementation adds `concourse-` by default to the namespace if it is not entered.

william-tran:
@pontusarfwedson `((common-secrets.api-key))` is the expression you use in your pipeline, so you can call the part before the dot whatever you want.

> I get the impression it only looks for the key value in pipeline+secret or just in secret.
The definitive description of its behaviour is this section here. https://github.com/kubernetes/charts/tree/master/stable/concourse#kubernetes-secrets

> Also, if we run our services in the namespace team1 and then have a team1 in concourse then we cannot make our secrets go to that namespace because the current implementation adds concourse- by default to the namespace if it is not entered.
If you set the namespace prefix to empty string, you can get the behaviour you're looking for.

pontusarfwedson:
Alright, thanks for the response. Just out of curiosity, is this achieved by the 
```
evenLessTyped := map[interface{}]interface{}{}
	for k, v := range secret.Data {
		evenLessTyped[k] = string(v)
	}

	return evenLessTyped, true, nil
```
functionality and resolved elsewhere? I wasn't sure what was going on there to be honest so got the impression it only tried to resolve the key called `value`.

william-tran:
> Because Concourse can schedule privileged containers being used by different teams beside eachother (e.g. to the same worker), it's conceivable that even if you didn't want secrets to be read across teams and namespaces, that you can't 100% prevent it the way it works today.

I take that back, workers can be registered for specific teams.

> is this achieved by the

I think so @pontusarfwedson... though I'm no pro at go by any means, it's similar logic to what we see in https://github.com/concourse/atc/blob/master/creds/vault/vault.go

vito:
Sorry, it's taking me a while to chew through PRs for our credential managers because I'm starting to see a trend between them that's making me think we should take a bigger step instead.

I've left a comment (https://github.com/concourse/concourse/issues/1814#issuecomment-392798692) on another Vault-related issue asking if anyone's interested in making RFCs for a "next iteration" of credential manager support. This PR feels pretty related but honestly I don't have the familiarity necessary with K8s secrets to be sure. After reading the linked comment do any of you (@pontusarfwedson @william-tran) feel that a new RFC for K8s is in order?

vito:
Gonna close this as it seems inactive now. I've opened https://github.com/concourse/rfcs/issues/5 per my recent comment.

PR #258, opened 2018-04-03
Add BuildLogRetentionCalculator

aeijdenberg:
Add `--default-build-logs-to-retain NN` and `--max-build-logs-to-retain NN` options to `atc` to specify a default and maximum number of build logs to retain.

The default is ignored if specified by a job, and the maximum (if specified) is always applied to cap the number of build logs retained.

Both default to `0`, which should have no effect on existing installations.

Addresses https://github.com/concourse/concourse/issues/1417

If this PR is accepted, I'll write another to add to the `bosh` job spec for `atc`.

### DIFF
--- a/gc/buildreaper_test.go
+++ b/gc/buildreaper_test.go
@@ -32,6 +32,7 @@ var _ = Describe("BuildReaper", func() {
 			buildReaperLogger,
 			fakePipelineFactory,
 			batchSize,
+			NewBuildLogRetentionCalculator(0, 0),
 		)
 	})
 
--- a/gc/buildreaper_test.go
+++ b/gc/buildreaper_test.go
@@ -19,11 +19,13 @@ var _ = Describe("BuildReaper", func() {
 		buildReaper         BuildReaper
 		fakePipelineFactory *dbfakes.FakePipelineFactory
 		batchSize           int
+		buildLogRetainCalc  BuildLogRetentionCalculator
 	)
 
 	BeforeEach(func() {
 		fakePipelineFactory = new(dbfakes.FakePipelineFactory)
 		batchSize = 5
+		buildLogRetainCalc = NewBuildLogRetentionCalculator(0, 0)
 	})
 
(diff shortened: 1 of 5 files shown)


gc/buildreaper_test.go
```diff
@@ -32,6 +32,7 @@ var _ = Describe("BuildReaper", func() {
 			buildReaperLogger,
 			fakePipelineFactory,
 			batchSize,
+			NewBuildLogRetentionCalculator(0, 0),
```
vito: Hmm I'm a little worried that the coverage here isn't sufficient. The build reaper could have been unchanged, either in the line checking if it's 0 or in the line adding it to `db.Page`, and all the tests would have passed.

Could tests be added here, either instead of or in addition to the tests for the calculator? (Depending on where the tests feel more natural to write, either with fakes or with a real object.)

aeijdenberg:
@vito - thanks for the review. I've added a small integration test to `buildreaper_test.go` that verifies that the `BuildLogRetentionCalculator` is used by the `BuildReaper`, and I've left the bulk of the calculation tests in the unit tests for the calculator itself.

Although the build claims to have failed, I think that's due to running out of space somewhere on the build server and unrelated to this change. I don't think I have the permissions to get it to kick-off a new one.

vito:
@aeijdenberg thanks! I'll kick off another build.

aeijdenberg:
Thanks @vito - build still seems to be broken though due to `No space left on device` in what appears to be an unrelated test.

vito:
@aeijdenberg Very strange - looking at the machine metrics, it's not running out of disk. I don't think it's related to your change at all, but I'm going to investigate a bit before merging, since we *should* be able to trust these build statuses.

vito:
Well, it passed this time. Hmm. Merging.

aeijdenberg:
Thanks @vito - I'll start work on a PR to add to the bosh release.

PR #259, opened 2018-04-03
Add WorkerName to ErrCreatedVolumeNotFound

rfliam:
The ErrCreatedVolumeNotFound is logged on the ATC, not the
workers. Determining which worker is failing is very difficult.  This
adds the worker name so operators can troubleshoot more quickly.


worker/volume_client.go
```diff
@@ -68,11 +68,12 @@ type VolumeClient interface {
...
 
 type ErrCreatedVolumeNotFound struct {
-	Handle string
+	Handle     string
+	WorkerName string
 }
 
 func (e ErrCreatedVolumeNotFound) Error() string {
-	return fmt.Sprintf("failed to find created volume in baggageclaim. Volume handle: %s", e.Handle)
+	return fmt.Sprintf("failed to find created volume in baggageclaim. Volume handle: %s WorkerName: %s", e.Handle, e.WorkerName)
```
vito: the format here is a bit odd (and was a bit odd before, too).

how about:

```
volume '%s' disappeared from worker '%s'
```

rfliam:
Agreed. I've update the message to this:
```volume 'fake-handle' disappeared from worker 'fake-worker' for pipeline 'fake-pipeline' job 'fake-job' task 'fake-task'```

Or this if we can't pull the details:
```volume 'fake-handle' disappeared from worker 'fake-worker', could not fetch task information err 'fake-error'```

File: worker/volume_client.go [hunk 2]
```diff
@@ -68,11 +68,15 @@ type VolumeClient interface {
 var ErrVolumeExpiredImmediately = errors.New("volume expired immediately after saving")
 
 type ErrCreatedVolumeNotFound struct {
-	Handle string
+	Volume db.CreatedVolume
 }
 
 func (e ErrCreatedVolumeNotFound) Error() string {
-	return fmt.Sprintf("failed to find created volume in baggageclaim. Volume handle: %s", e.Handle)
+	p, j, t, err := e.Volume.TaskIdentifier()
```
vito: Hmm I don't think the pipeline/task information will always be correct, since this error could happen for other reasons too (like checking resources or running a `get`/`put` step).

The message should already be in the appropriate context, since the UI will show the error in the output of the step that was running, and the system log lines probably have enough context in the message/session/data.

I'd recommend keeping it shorter and only showing the more critical information.

PR #262, opened 2018-04-10
Credentials provider based on the newly announced AWS SecretsManager

x6j8x:
https://aws.amazon.com/documentation/secretsmanager/

- supports multi-value variables by storing JSON strings as SecretBinary (Example: `{"user": "user", "password": "password"}`)
- single value entries must be stored as SecretString
- retrieves secrets in the default AWSCURRENT staging label

atc must be granted the "secretsmanager:GetSecretValue" permission to be able to read secrets.

requires aws-go-sdk to be updated to at least v1.13.28 (in github.com/concourse/concourse)

### DIFF
--- a/atccmd/command.go
+++ b/atccmd/command.go
@@ -72,6 +72,7 @@ import (
 	// dynamically registered credential managers
 	_ "github.com/concourse/atc/creds/credhub"
 	_ "github.com/concourse/atc/creds/kubernetes"
+	_ "github.com/concourse/atc/creds/secretsmanager"
 	_ "github.com/concourse/atc/creds/ssm"
 	_ "github.com/concourse/atc/creds/vault"
 )
--- a/creds/secretsmanager/manager.go
+++ b/creds/secretsmanager/manager.go
@@ -1,0 +1,109 @@
+package secretsmanager
+
+import (
+	"errors"
+	"io/ioutil"
+	"text/template"
+	"text/template/parse"
+
+	"code.cloudfoundry.org/lager"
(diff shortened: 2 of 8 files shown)


pivotal-issuemaster:
@x6j8x Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=262)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/atc&pullRequestId=262) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@x6j8x Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/atc&pullRequestId=262)!

vito:
Awesome, thanks for putting this together! It looks like it's mostly based on the existing SSM backend, which makes sense. I have a hunch that most people should migrate to this new service and perhaps we'll be able to retire the old one in the future (however distant it may be).

I'll merge it locally and bump the `aws-sdk-go` submodule.

x6j8x:
Awesome! @vito yes it’s largely based on the ssm provider. Only the map[] handling is different (because there’s no queryByPath in SecretManager).

zekizeki:
I wouldn't retire the SSM without a close look. SSM is still a valid choice for secrets management. It has the advantage of not having a per credential charge. However it doesn't support credential rotation and has tighter API rate limiting than AWS secrets manager. But is still a valid choice.

PR #266, opened 2018-04-18
Support migration options in web binary

SHyx0rmZ:
This moves the logic to handle migration commands from
(*atccmd.ATCCommand).Execute() to (*atccmd.ATCCommand).Runner(),
so that in case of the migration options being present, a new runner
is returned, which will only handle the migration.

Closes concourse/concourse#2163.


vito:
thanks!

PR #267, opened 2018-04-18
If a session signing key is generated, store it in the command

SHyx0rmZ:
In case the user doesn't provide a session signing key, we will
generate one on the fly. Instead of just returning the generated
key, we should also store it in the command, as the key in the
command is referenced in multiple other places.

Closes concourse/concourse#2152.

### DIFF
--- a/atccmd/command.go
+++ b/atccmd/command.go
@@ -981,20 +981,16 @@ func (cmd *ATCCommand) constructWorkerPool(
 }
 
 func (cmd *ATCCommand) loadOrGenerateSigningKey() (*rsa.PrivateKey, error) {
-	var signingKey *rsa.PrivateKey
-
 	if cmd.SessionSigningKey.PrivateKey == nil {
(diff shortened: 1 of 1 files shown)


SHyx0rmZ:
Ha, I didn't think of any ordering requirements. Yep, that sounds better, will update the PR.

SHyx0rmZ:
Shouldn't we really add that logic to `concourse/tsa` instead of `concourse/bin`?

vito:
@SHyx0rmZ Yeah, the TSA should just blow up if it's not given a key - it's the client, not the thing verifying the key, so if it were to generate it then it'd just be making tokens that no one can verify.

The BOSH release doesn't invoke `bin` but it doesn't need the keys generated on start in the first place - that's typically done via variables in the manifest, which BOSH can generate for you.

vito:
Closing in favor of https://github.com/concourse/bin/pull/43

PR #268, opened 2018-04-19
Release the lock in case of an error

SHyx0rmZ:
I'm hoping this might fix concourse/concourse#1954.


vito:
Wow, good find. Thanks!

PR #270, opened 2018-04-22
[WIP] Create credhub client lazily.

jmcarp:
Creating a credhub client can fail if credhub isn't available
(specifically if we pass options that require connecting to credhub). To
allow the atc to start up when credhub is unavailable, create a lazy
credhub client that doesn't talk to credhub until the credhub provider
actually needs to look up credentials.

Not sure where to add tests--it looks like the credhub backend doesn't have any yet. WDYT @vito ?

[Resolves #2154]

### DIFF
--- a/creds/credhub/credhub.go
+++ b/creds/credhub/credhub.go
@@ -11,7 +11,7 @@ import (
 )
 
 type CredHubAtc struct {
-	CredHub *credhub.CredHub
+	CredHub lazyCredhub
 	logger  lager.Logger
 
 	PathPrefix   string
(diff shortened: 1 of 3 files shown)


vito:
Makes sense to me. I don't think there are any unit-level tests here, but we cover CredHub integration in TopGun so I'm not too worried. In fact I wouldn't be too shocked if the logs during TopGun show that the ATC fails to start until CredHub is up.

I think I can just merge this as-is if the WIP was just for tests. Sound good?

jmcarp:
Yep, I just marked this WIP for lack of tests.

vito:
Thanks!

PR #273, opened 2018-05-05
Allow setting full credhub paths.

jmcarp:
At the moment, credhub credential lookup requires keys to be prefixed with either `/$prefix/$team/$pipeline/` or `/$prefix/$team/`. That works all right when concourse is the only consumer of credhub secrets, but it's not great for sharing credentials with another system, like bosh. For example, I want to create a uaa client in a bosh deployment, and then use that client in a concourse pipeline. By default, that secret would live at `/$director/$deployment/$secret`, so I can't look it up from a concourse pipeline. It would be useful if concourse pipelines could look up credentials with bosh prefixes, or if I could store those credentials at a different path, like `/common/clients/`.

I think the bosh credhub integration has a reasonable solution for this problem: prefix variables with reasonable defaults unless the user specifies a full path: https://github.com/cloudfoundry-incubator/credhub/blob/master/docs/operator-quick-start.md#variable-namespacing. I'm proposing we do the same thing here. Arguably, this exposes too many secrets to every pipeline, but credhub authorization should really belong to credhub, not to concourse. Ideally, every team would get different credhub client credentials: https://github.com/concourse/concourse/issues/1814#issuecomment-377778487. That would be a bigger refactor--for now, we might say that only the main team can look up credentials by full path.

WDYT @vito?

### DIFF
--- a/creds/credhub/credhub.go
+++ b/creds/credhub/credhub.go
@@ -2,6 +2,7 @@ package credhub
 
 import (
 	"path"
+	"strings"
 
 	"code.cloudfoundry.org/lager"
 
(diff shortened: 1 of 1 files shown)


vito:
I totally see why you'd want this but I'm really not comfortable with permitting this even as an interim solution. I don't like the idea of our CredHub integration being less restrictive/secure than the others, even if it's not a big deal for most/all use cases of CredHub. We also can't really know how people are using Concourse + CredHub, and given the nature of this feature it'd be *basically* the worst thing in the world if we suddenly permitted teams to access each others' credentials just by guessing names.

I'm gonna close this and leave a comment on concourse/concourse#1814 to gauge interest in a new RFC for the next iteration of credential manager support. See ya there, and sorry it took me so long to reply to this!

PR #274, opened 2018-05-17
prometheus: Remove job scheduling metric

databus23:
For each pipeline the `job_scheduling_duration` metric is emitted for each job in that pipeline during each scheduling tick. Without this label the prometheus metric is not really helpful as it only reflects the duration of an arbitrary job that was emitted last before the scrape.

/cc @TimSimmons

### DIFF
--- a/metric/emitter/prometheus.go
+++ b/metric/emitter/prometheus.go
@@ -28,10 +28,8 @@ type PrometheusEmitter struct {
 
 	httpRequestsDuration *prometheus.HistogramVec
 
-	schedulingFullDuration        *prometheus.GaugeVec
-	schedulingLoadingDuration     *prometheus.GaugeVec
-	schedulingJobDurationTotalVec *prometheus.CounterVec
-	schedulingJobTotalVec         *prometheus.CounterVec
+	schedulingFullDuration    *prometheus.GaugeVec
+	schedulingLoadingDuration *prometheus.GaugeVec
 
 	dbQueriesTotal prometheus.Counter
 	dbConnections  *prometheus.GaugeVec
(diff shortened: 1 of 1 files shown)


TimSimmons:
I can what you're getting at here, I wonder if it might be more useful if these scheduling duration metrics were a Histogram? That way you'd be able to get an idea across all the instances of the event, vs just the last one.

My other concern with this is the metric cardinality this will result in. I won't pretend I'm intimately familiar with the internals here, but is it useful to know how long a certain job in a pipeline takes to schedule? What makes jobs within a pipeline different as far as these scheduling durations go?

databus23:
Regarding the histogram idea, I don't think thats needed in this case. Each pipeline is processed in its own loop with a sleep of 10 seconds. So the events for a pipeline/job are emitted sequentially at a rate not faster then 6 times per minute. I think its more helpful to have exact durations for the sampled scheduling ticks than a histogram providing less precision and a higher metric cardinality.

Regarding the overall cardinality it highly depends on the size of the concourse deployment (number of pipelines and jobs per pipeline) but it could easily by hundreds of metrics. Nonetheless I would prefer to have these metrics available.
We are facing serious performance problems with our rather large concourse deployment for some time now and are still trying to understand which pipelines are causing this.
As I am understanding this the time it takes to schedule an individual pipeline job depends mainly on how many versions are recorded for the individual resources that go into the job.

We for example had a poorly written custom resource that by accident discovered a new version during each `check` run (every 30 seconds). Because we didn't trigger on the resource we didn't notice it at first. After a few month that resources had 300k versions in the databasae and all jobs using that resource took several minutes to schedule.

vito:
Hmm, I also share @TimSimmons's concerns over cardinality here. These would definitely be valuable in understanding why your Concourse is falling over, but that may in turn make peoples' InfluxDB fall over instead.

Could you instead use the logging-based metrics emitter for debugging this? That won't work right now because I'm pretty sure we only allow one emitter to be configured, but that could be an easy change. You could then configure `--prometheus-...` and `--emit-to-logs`, using your log sink to view higher-fidelity metrics like per-job scheduling times, without making Prometheus unhappy in large multi-tenant installations.

databus23:
@vito This change is not adding any new metrics. Its just exposing the metrics that are already emitted  with the correct label set to prometheus. the Influxdb emitter already sends this out periodically for every job.

Do we have any numbers available about how many distinct jobs are defined in a large concourse installation so that we can reason about the cardinality?

databus23:
@TimSimmons speaking of cardinality: I double that with the last commit by switching from a single gauge per job to two counters. I noticed that gauges have to problem that you can't really notice when a metric is going away (e.g. job is deleted). the eporter will happily serve the last emitted value until the application is restarted.

In general I would suggest to switch to counters for the other pipeline scheduling metrics (`duration_full` and `duration_loading`) so that deleted pipelines are properly reflected in the graphs.

databus23:
Hmm out postgresql database has 2284 rows in the jobs table, that could be a bit much.

databus23:
@vito Wasn't me :) I just finally got around enabling metrics using the prometheus emitter. I tried getting from riemman to prometheus but gave up.

Ok but that probably means we can't expose the job specific metrics in prometheus and should remove the metrics that is currently there as its bogus.

databus23:
@vito @TimSimmons Should I update the pull request to remove the job scheduling metric then? The current implementation is misleading and adding the appropriate job label would increase the metric cardinality significantly.

vito:
@databus23 Yeah, I think that's a good idea if the data it's emitting is impossible to consume usefully.

databus23:
@vito @TimSimmons Ok I updated the PR to remove the job scheduling metric for now. PTAL

PR #275, opened 2018-05-17
Add platform label to worker metrics

databus23:
The number of volumes/containers varies greatly per platform. In order to detect anomalies (e.g. this linux worker has to few/many volumes/containers) I think adding the platform as a label makes sense.


vito:
Seems nifty. There are only a few platforms so I'm not very worried about cardinality here. Thanks!

PR #276, opened 2018-05-17
prometheus: Use counter for db queries

databus23:
This renames the concourse_db_queries metric to concourse_db_queries_total and changes it to a counter that tracks the total number of db queries since application start.

This is way more accurate and ideomeatic for prometheu's pull model. The previous approach was very error prune and almost certainly very incorrect.

@TimSimmons Any comments?

### DIFF
--- a/metric/emitter/prometheus.go
+++ b/metric/emitter/prometheus.go
@@ -31,6 +31,9 @@ type PrometheusEmitter struct {
 	schedulingFullDuration    *prometheus.GaugeVec
 	schedulingLoadingDuration *prometheus.GaugeVec
 	schedulingJobDuration     *prometheus.GaugeVec
+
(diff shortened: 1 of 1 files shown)


TimSimmons:
It's been a while, but I believe I used `ConstMetrics` (which are [generally the more idiomatic method](https://prometheus.io/docs/instrumenting/writing_exporters/#collectors) for this type of thing) for a reason.

But now that I look at it again I think that reason might have been that I didn't know they were periodically emitted: https://github.com/concourse/atc/blob/d3cc08277460b8dbd220862d32a56eb7625581ac/metric/periodic.go
So this is probably a perfectly fine way to do things :+1: and will give you a nice counter for database queries over time.

vito:
thanks!

PR #277, opened 2018-05-30
prometheus: Remove stale worker metrics

databus23:
This fixes a bug reported in https://github.com/concourse/concourse/issues/2244

It removes gauge metrics for workers that haven't received a metric event in 5 minutes (because they are stalled or retired).

/cc @TimSimmons @jmcarp

### DIFF
--- a/metric/emitter/prometheus.go
+++ b/metric/emitter/prometheus.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"sync"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/atc/db"
@@ -35,6 +37,9 @@ type PrometheusEmitter struct {
 	dbConnections  *prometheus.GaugeVec
 
 	resourceChecksVec *prometheus.CounterVec
+
+	workerLastSeen map[string]time.Time
+	mu             sync.Mutex
 }
 
 type PrometheusConfig struct {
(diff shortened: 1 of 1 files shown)


databus23:
I updated the PR to only handle stale worker metrics for now. Thinking about this some more I think I would rather change the scheduling metrics to counters which are not as problematic as gauges when the time series does not get anymore updates (because the pipeline is paused or deleted). I will create a separate PR for that.

metric/emitter/prometheus.go
```diff
@@ -421,3 +435,33 @@ func (emitter *PrometheusEmitter) databaseMetrics(logger lager.Logger, event met
...
+func (emitter *PrometheusEmitter) periodicMetricGC() {
+	for {
+		emitter.mu.Lock()
+		now := time.Now()
+		for worker, lastSeen := range emitter.workerLastSeen {
+			if now.Sub(lastSeen) > 5*time.Minute {
+				//this is a little stupid but we don't know the platform here,
+				//so we try all known platoforms
+				for _, platform := range []string{"linux", "windows", "darwin"} {
+					emitter.workerContainers.DeleteLabelValues(worker, platform)
```
vito: Does this need the platform? Shouldn't the worker name be enough? Those are globally unique within Concourse's database.

databus23: The problem is that `DeleteLabelValues` only removes timeseries that have an exact match ok the label set. 

See https://godoc.org/github.com/prometheus/client_golang/prometheus#GaugeVec.DeleteLabelValues

vito: Gotcha, so it runs against all platforms just to cover all the bases. Wish there was a better way to do that, but it makes sense.

Unfortunately there are conflicts now so I'll hold off for a rebase, but I'll approve the changes for now.

databus23:
@vito squashed and rebased.

vito:
thanks!

PR #278, opened 2018-06-01
prometheus: Improve label consistency

databus23:
This PR actually fixes two problems:

* Add worker platform attribute during initial registration
* Don't add time series to prometheus when labels are missing


vito:
thanks!

PR #279, opened 2018-06-04
Add resource check metric

databus23:
This PR adds a metric event whenever a resource check is performed.

I think its valuable to have insights into the continuous load that is caused by the periodical resource checks that are not really reflected in the existing metrics.

For example we are seeing patterns where users create pipelines with a lot of git resources using the `path` filter for a large monorepos. This causes a lot of load on the workers (performing the checks) and also on outside resources (git repository server). Relating to https://github.com/concourse/concourse/issues/2249 I think its valuable to identify changes to pipelines (or new pipelines) that cause a spike in resource checks.

While creating this PR I wondered this rather should be be a periodically emitted counter (per pipeline) to avoid sending to much events? Assuming the resource default configuration this will emit an event about every 30 seconds for each resource. In comparison the job scheduling metrics are emitted for every active job every 10 seconds which is roughly in the same ball park.

### DIFF
--- a/metric/emitter/prometheus.go
+++ b/metric/emitter/prometheus.go
@@ -33,6 +33,8 @@ type PrometheusEmitter struct {
 
 	dbQueriesTotal prometheus.Counter
 	dbConnections  *prometheus.GaugeVec
+
+	resourceChecksVec *prometheus.CounterVec
 }
 
 type PrometheusConfig struct {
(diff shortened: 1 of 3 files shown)


vito:
Neat, thanks! This should be useful to monitor on large environments.

marco-m:
Thanks @databus23 we are suffering from the git resources check frequency, this metrics will be very useful!

databus23:
@marco-m Who doesn't suffer from the git resource check frequency? 😆

PR #282, opened 2018-06-09
ATCCommand now tells callers to skip the TSA for migration commands

SHyx0rmZ:
This should fix concourse/concourse#2261.


vito:
Thanks! I'm OK with this approach for now but we should probably just have a separate command or something in the future.

SHyx0rmZ:
Do you have any name in mind for that command? Maybe `database`?

vito:
How about we split the flags into separate commands?

```sh
concourse current-db-version
concourse supported-db-version
concourse migrate --version=<version> # required flag
```

This could be used like so (for a downgrade):

```sh
new-concourse migrate --version=$(old-concourse current-db-version)
```

PR #283, opened 2018-06-18
prometheus: Convert scheduling metrics from gauges to counters

databus23:
Instead of sampling the scheduling pipeline metrics at an arbitray rate (depending on the prometheus scrape interval) we now have a counter summing up the time it took to schedule a pipeline. The additional counter (`concourse_scheduling_total`) can be used to calculate an average scheduling time within a scrape interval.

Using counters for the scheduling metrics has the added benefit that it handles deleted or paused pipelines better. Previously the scheduling time was stuck at the value that was emitted last. Now the rate of change drop to zero clearly indicating that the pipeline is not beeing scheduled anymore.

Note: This change is breaking, meaning folks using the prometheus exporter will need to updated their (scheduling) graphs.

/cc @TimSimmons @vito

### DIFF
--- a/metric/emitter/prometheus.go
+++ b/metric/emitter/prometheus.go
@@ -30,8 +30,9 @@ type PrometheusEmitter struct {
 
 	httpRequestsDuration *prometheus.HistogramVec
 
-	schedulingFullDuration    *prometheus.GaugeVec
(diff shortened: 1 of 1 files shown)


vito:
Thanks!

PR #284, opened 2018-06-22
optimize versions db cache busting

vito:
store an integer column on the pipeline, bumped every time the versions
db should be invalidated (new versioned resources, new build inputs, new
build outputs, build finishes)

the old query was insanely slow when run against larger data sets, and
required a ton of bookkeeping (modified_time on 3 tables). this should
be much, much faster.

thankfully there are high-level tests for this behavior, so it was pretty easy
to validate that i'm bumping the index in all the right places

### DIFF
--- a/db/build.go
+++ b/db/build.go
@@ -334,6 +334,13 @@ func (b *build) Finish(status BuildStatus) error {
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


xtremerui:
Looks good to me!

PR #290, opened 2018-07-12
[#35] Export Teams from Concourse Instance

pivotal-saman-alvi:
* Add `auth` details to `teams` api reponse

This was done to expose the authentication information set for a team, completed as part of the feature to be able to set multiple teams.

In conjunction with PR: https://github.com/concourse/fly/pull/240

Signed-off-by: Saman Alvi <salvi@pivotal.io>

### DIFF
--- a/api/present/team.go
+++ b/api/present/team.go
@@ -9,5 +9,6 @@ func Team(team db.Team) atc.Team {
 	return atc.Team{
 		ID:   team.ID(),
 		Name: team.Name(),
+		Auth: team.Auth(),
 	}
 }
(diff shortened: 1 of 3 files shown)


pivotal-jwinters:
Maybe we shouldn't show you the teams at all if you're not a member/admin?

vito:
@pivotal-jwinters yeah good point, now that the auth is by user we don't need to list them all

pivotal-saman-alvi:
Modified to only show teams which the user has access to @pivotal-jwinters @vito

vito:
Thanks all!

PR #291, opened 2018-07-13
Fix issue #2300. lazyCredhub.Credhub() did not receive a pointer

iferunsewe:
This PR allows the ATC to start if CredHub is down, without spending lots of CPU time instantiating objects.

The previous attempts to fix this had a value receiver, meaning that each time the `CredHub()` method was invoked, `lc` was copied to the stack, had its `credhub` field set, and then the instance of `lc` was discarded when the method returned and we pop off the stack. This meant that `lc.credhub` was _always_ nil, and so we _always_ constructed a new `credhub.CredHub` instance.

We opted to use a mutex across the whole method, because the previous implementation that only ran once would mean that if the construction of `credhub.CredHub` returned an error, there would never be an opportunity to try again - effectively the 'once' bit of it wasn't dependent on a successful outcome.

### DIFF
--- a/creds/credhub/manager.go
+++ b/creds/credhub/manager.go
@@ -107,10 +107,8 @@ type lazyCredhub struct {
 	url     string
 	options []credhub.Option
 	credhub *credhub.CredHub
+	mu      sync.Mutex
 }
-
-var credhubInstance *credhub.CredHub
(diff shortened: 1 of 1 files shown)


vito:
Good catch! Thanks!

PR #294, opened 2018-07-27
use ON CONFLICT for upserts in more areas

vito:
...instead of swallowing unique violation and foreign key errors and
retrying.

the approach here is two-fold:

* instead of selecting and then handling a foreign key error on insert,
  select FOR SHARE so that it can't disappear while the transaction is
  using it
* instead of inserting and then handling a unique violation error, use
  `ON CONFLICT DO UPDATE SET ...`


vito:
Turns out this should fix https://github.com/concourse/concourse/issues/2454! But we should add a test now that we know an exact scenario to cover.

PR #296, opened 2018-08-09
Validate interpolated variables during fly set-pipeline

edtan:
This checks incoming pipeline configuration from `fly set-pipeline` and checks to see if all credentials exist in a credential manager, as requested in concourse/concourse#1920.  So far, I've only tested this with Vault.  Note that this does not currently support validating credentials in a [task file](https://concourse-ci.org/task-step.html#file).

I have left some questions/comments in the code for review.  This also adds a `--skip-creds` option to `fly set-pipeline` to skip this validation, in case operators might want to override the check for various reasons.  This is intended to go with
concourse/go-concourse#15
concourse/fly#252


edtan:
I've finished up the changes to go-concourse and fly to add a `skip-creds` option to disable the credentials check.

Also, in order to support the `skip-creds` option, I added a new `configskipcredentials` endpoint, but I don't know if this is a good way to do it?  As I was writing tests, it seems like I had to write a lot of duplicate tests.

api/configserver/save.go
```diff
@@ -50,6 +52,14 @@ type SaveConfigResponse struct {
 }
 
 func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
+	s.saveConfig(w, r, true)
+}
+
+func (s *Server) SaveConfigSkipCredentials(w http.ResponseWriter, r *http.Request) {
```
vito: Could this be a query param instead of a separate API endpoint? And maybe have it so you opt-in to credential checking rather than doing it by default? Something like `?check_creds`.

api/configserver/save.go
```diff
@@ -145,6 +168,71 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 	s.writeSaveConfigResponse(w, SaveConfigResponse{Warnings: warnings}, session)
 }
 
+// Simply validate that the credentials exist; don't do anything with the actual secrets
+func validateCredParams(vars creds.Variables, config atc.Config, session lager.Logger) []string {
+	// 1. currently doesn't support tasks specified using a task file
+	// 2. should we check each of these are not nil first before Evaluating them to prevent unnecessary lookups, or is this already cached in vars?
+
+	errs := []string{}
```
vito: Could be a use case for `multierror`.

api/configserver/save.go
```diff
@@ -145,6 +168,71 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
...
+
+	for _, resourceType := range config.ResourceTypes {
+		_, err := creds.NewSource(vars, resourceType.Source).Evaluate()
+		if err != nil {
+			session.Debug(err.Error())
```
vito: Let's clean up these debug lines - can we just have one log line at the end with the aggregate of the errors?:

Something like:

```go
if err != nil { // assuming use of mutlierror.Append
  session.Info("config-has-invalid-creds", lager.Data{"errors": err.Error()})
}
```

api/configserver/save.go
```diff
@@ -145,6 +168,71 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
...
+
+				_, err = creds.NewTaskParams(vars, plan.TaskConfig.Params).Evaluate()
+				if err != nil {
+					session.Debug(err.Error())
+					errs = append(errs, err.Error())
+				}
+
+				// Are ImageResource.Params needed too?  According to the docs at https://concourse-ci.org/creds.html#vault,
+				// it seems like they aren't supported at the moment
+				// _, err = creds.NewParams(vars, plan.TaskConfig.ImageResource.Params).Evaluate()
```
vito: Yeah, I don't think we support evaluating `image_resource.params`. Support for `params` on `image_resource` was added for non-credentialy things. I'd expect all creds to be in `source`.

api/handler.go
```diff
@@ -94,8 +94,9 @@ func NewHandler(
 	infoServer := infoserver.NewServer(logger, version, workerVersion, credsManagers)
 
 	handlers := map[string]http.Handler{
-		atc.GetConfig:  http.HandlerFunc(configServer.GetConfig),
-		atc.SaveConfig: http.HandlerFunc(configServer.SaveConfig),
+		atc.GetConfig:                 http.HandlerFunc(configServer.GetConfig),
+		atc.SaveConfig:                http.HandlerFunc(configServer.SaveConfig),
+		atc.SaveConfigSkipCredentials: http.HandlerFunc(configServer.SaveConfigSkipCredentials),
```
vito: (Already mentioned above but I think this should just be a query param on the existing endpoint.)

edtan:
I made the validation turned off by default as you suggested.  I think it makes more sense, because as one is configuring a pipeline, they probably don't want to hit the credential manager until they've finalized the configuration.

I'll also take a look into that `List()` method.  This might take me a while as I'm not familiar with credential managers other than Vault.

api/configserver/save.go
```diff
@@ -106,11 +116,24 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	session.Info("saving")
-
 	pipelineName := rata.Param(r, "pipeline_name")
 	teamName := rata.Param(r, "team_name")
 
+	if checkCredentials {
+		variables := s.variablesFactory.NewVariables(teamName, pipelineName)
+		_, isNoop := variables.(*noop.Noop)
```
vito: Hmm I don't think we need this special case. I could see a user still wanting to check that all vars resolve, and expect this to fail, since in reality their builds will fail. That way the user can know that the vars aren't satisfied by a remote credential manager, so they need to provide them locally with `-v`/`-y`/`-l`.

edtan: Ah yes, that makes sense.  The `Noop` special case is now gone.

vito:
Awesome, thanks!

PR #297, opened 2018-08-11
Validate that params are declared in task file before merging

edtan:
For tasks defined by `file: task.yml`, if params are defined in the outer task step, but not declared within the task.yml file, return an error.  This addresses concourse/concourse#555.

It would also be nice to update `fly set-pipeline` with the same type of validation for faster feedback, but I don't know how to implement that.

### DIFF
--- a/exec/task_config_source.go
+++ b/exec/task_config_source.go
@@ -177,7 +177,7 @@ func (configSource MergedConfigSource) FetchConfig(source *worker.ArtifactReposi
 		return atc.TaskConfig{}, err
 	}
 
-	return aConfig.Merge(bConfig), nil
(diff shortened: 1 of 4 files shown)


vito:
This is great, thanks!

I don't think we need the validation in `set-pipeline` since the task config is fetched at runtime. It's also possible that a pipeline is "valid" at configuration time but becomes invalid when someone changes the task config and pushes. So it makes more sense to just throw a runtime error.

vito:
We're gonna turn this into a warning instead for now, since otherwise this is backwards-incompatible. We'll turn it back into an error with 5.0.

PR #298, opened 2018-08-13
Add API to support clearing of task caches from fly

edtan:
Add API to support clearing of task caches from fly.  Supports concourse/concourse#1817.

Intended to go with:
concourse/go-concourse#16
concourse/fly#254

### DIFF
--- a/api/jobserver/clear_task_cache.go
+++ b/api/jobserver/clear_task_cache.go
@@ -8,6 +8,7 @@ import (
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/atc"
 	"github.com/concourse/atc/db"
+	"github.com/google/jsonapi"
 )
 
 func (s *Server) ClearTaskCache(pipeline db.Pipeline) http.Handler {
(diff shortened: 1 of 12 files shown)


db/job.go
```diff
@@ -581,6 +583,39 @@ func (j *job) CreateBuild() (Build, error) {
 	return build, nil
 }
 
+func (j *job) ClearTaskCache(stepName string, cachePath string) (int64, error) {
+	tx, err := j.conn.Begin()
+	if err != nil {
+		return -1, err
```
vito: It's more common in Go to return zero-values in err cases (so `0` instead of `-1`).

db/job.go
```diff
@@ -581,6 +583,39 @@ func (j *job) CreateBuild() (Build, error) {
...
+	}
+
+	defer Rollback(tx)
+
+	var sqlBuilder sq.DeleteBuilder
+	sqlBuilder = psql.Delete("worker_task_caches").
+		Where(sq.Eq{"job_id": j.id})
```
vito: `stepName` should be part of this query

db/job.go
```diff
@@ -581,6 +583,39 @@ func (j *job) CreateBuild() (Build, error) {
...
+	if len(cachePath) > 0 {
+		sqlBuilder = sqlBuilder.Where(sq.Eq{"path": cachePath})
+	}
+
+	sqlResult, err := sqlBuilder.
+		RunWith(tx).
+		Exec()
+
+	if err != nil {
+		return -1, err
```
vito: Should be `return 0, err` per prior comment

job.go
```diff
@@ -43,3 +43,7 @@ type BuildInput struct {
 	Version  Version  `json:"version"`
 	Tags     []string `json:"tags,omitempty"`
 }
+
+const (
+	ClearTaskCacheQueryPath = "cachePath"
```
vito: Hmm this might make more sense in `routes.go`

api/jobserver/clear_task_cache.go
```diff
@@ -0,0 +1,76 @@
...
+
+		rowsDeleted, err := job.ClearTaskCache(stepName, cachePath)
+
+		if rowsDeleted == 0 {
+			logger.Debug("could-not-find-cache-path", lager.Data{
+				"jobName":   jobName,
+				"stepName":  stepName,
+				"cachePath": cachePath,
+			})
+			w.WriteHeader(http.StatusNotFound)
```
vito: Hmm, what do you think about having this not be an error case and instead just return the number of caches that were deleted? Clearing caches is something that could easily be idempotent, but I can see why you'd want to surface *something* as the user may have typo'd the path or step name or something. Showing how many were deleted would let the user notice if they made a typo (because they'd expect a nonzero amount) or reason that maybe the cache actually wasn't there themselves. (It'd also be a nice confirmation to see how many were cleared.)

edtan: Ah yes, I was a bit uncertain about returning an error here.  I think it's a great idea to just return the number of deleted caches - it nicely handles both cases where either the caches were successfully deleted or the user may have made a typo.  Will update the code with these changes.

edtan:
No prob!  I've made the requested changes.  Also, I wasn't sure where to put the new `ClearTaskCacheResponse` struct, but I've put it in `task.go` for now.

task.go
```diff
@@ -10,6 +10,11 @@ import (
 	"github.com/mitchellh/mapstructure"
 )
 
+// Not sure where best to put this?
```
vito: Hmm, maybe we can just make a new `responses.go` file for these. It looks like there are a couple others that could be moved there too, though that's not necessary as part of your PR (`configserver.SaveConfigResponse` and `atc.ConfigResponse`).

routes.go
```diff
@@ -28,7 +28,8 @@ const (
 	JobBadge       = "JobBadge"
 	MainJobBadge   = "MainJobBadge"
 
-	ClearTaskCache = "ClearTaskCache"
+	ClearTaskCache          = "ClearTaskCache"
+	ClearTaskCacheQueryPath = "CachePath"
```
vito: This should be `"cache_path"` since that's how we format all the other query params.

task.go
```diff
@@ -10,6 +10,11 @@ import (
...
+type ClearTaskCacheResponse struct {
+	NumCacheRemoved int64 `json:"num_cache_removed"`
```
vito: Can we call this `caches_removed`? Just reads a bit nicer imo.

task.go
edtan: Sounds good.  I've only moved the `ClearTaskCacheResponse` struct to the `responses.go`, but might move the other responses in a separate PR.

vito:
Thanks a lot! Merged it manually with a commit on top to change the query param (minor consistency nitpick). I also opened https://github.com/concourse/concourse/issues/2517 fyi for additional acceptance test coverage for this, which we're happy to do ourselves or help you out if you wanna do it yourself.

PR #301, opened 2018-08-25
Change error to warning for task params defined in pipeline but not task.yml

edtan:
This PR changes an error to a warning for concourse/concourse#2511.

While working on this, I noticed that `StaticConfigSource` and `FileConfigSource` currently return an empty string slice.  I didn't make any changes to them similar to how I changed `MergedConfigSource`, but let me know if I should update them as well for consistency.

### DIFF
--- a/exec/task_config_source.go
+++ b/exec/task_config_source.go
@@ -3,7 +3,6 @@ package exec
 import (
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"math"
 	"strconv"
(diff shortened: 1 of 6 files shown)


exec/task_config_source.go
```diff
@@ -84,15 +84,11 @@ func (configSource DeprecationConfigSource) FetchConfig(repo *worker.ArtifactRep
 		return atc.TaskConfig{}, err
 	}
 
-	for _, warning := range configSource.Delegate.Warnings() {
```
vito: We could probably just remove the entire `DeprecationConfigSource` type at this point. I think its entire purpose was just to print the warnings, but it was weird to have it be itself an implementation of the interface. Printing them in the task step makes a lot more sense, and with the printing removed here there's really nothing that this object does anymore. (I don't think we even use it anywhere, come to think of it.)

vito:
Thanks! Looks like there's a build failure though: https://ci.concourse-ci.org/builds/2456281#L5b80aa27:50

edtan:
I've removed `DeprecationConfigSource` and its corresponding tests (they seemed to be duplicates of tests for `StaticConfigSource`).

Cool! I'll update my Go to 1.11.

vito:
Thanks! Merging.

PR #302, opened 2018-09-07
Remove GC Concurrency limits across worker pool

topherbullock:
smaller change for now is to remove guardrails from job runner
but keep its worker sync-ing to keep the list of running workers
up to date.

Worker Pool just calls Worker Provider to get the list of running 
workers, so it can be simplified if we use Worker Provider directly 
instead.


vito:
thanks!

